PR #12, opened 2021-05-23
Updated docs and sidebar

AdamCroply:
- Renamed old docs folder to docs-ref for storage of reference documentation.
- Created new docs folder and created starter pages for content creation
- Structured starting nav items for sidebar menus


jlvandenhout:
I think we should do without the numbering in the folder structure as it gets quite tedious to maintain if we want for example to add content in between and these the names of the folders and files result in ids you can link to, so you would have to make sure to update the numbering everywhere.

I would suggest changing the naming scheme of the folders and files to short, dash separated lowercase names, like: docs/getting-started/introduction.md. So we can easily link them. Then if we want more fine grained control of placement in the folder structure or the title displayed in the navbar, we can control that per file in the preamble or per folder in the _category_.json.

See [https://github.com/facebook/docusaurus/tree/master/website/docs](the docusaurus repo) for what this would look like.

jlvandenhout:
Just FYI, the rendering system defaults to alfabetical ordering and uses the first title in the document as menu item text.

jlvandenhout:
Other than that I think this is a good structure. If we can resolve the file naming, we can merge.

AdamCroply:
Ok, that makes a bit of sense, but how do we go about restructuring the content order within the folders, as we would naturally require a better flow of content through using our own order (rather than alphabetical)

AdamCroply:
These should be all updated now. Refactored the naming system for the docs and folders, and fixed the docusaurus markdown formatting.

PR #21, opened 2021-05-25
Add workflow to deploy to dev server on push to develop

jlvandenhout:
We have set up a dev server with restricted access. This workflow will deploy the development branch to the dev server on every push, so the latest changes can be viewed by those who have access and are not able to checkout and build locally.

This workflow requires a few secrets to set up. I will do this once you agree on this approach.

### DIFF
--- a/.github/workflows/dev-server-deploy.yml
+++ b/.github/workflows/dev-server-deploy.yml
@@ -1,0 +1,32 @@
+name: Dev Server Deploy
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


Phyloiota (commented):
I cannot figure out the code :) But i agree with the idea of the dev server and think thats a great idea to have an ever online active state of the project available. Any secrets that are needed should be used...

AdamCroply (approved):

PR #32, opened 2021-06-05
Add new wiki structure

Dr-Electron:
# Description of change

I added the new wiki structure draft. 

## Type of change

- Documentation Fix

## Change checklist


docusaurus.config.js
```diff
@@ -18,65 +18,135 @@ module.exports = {
       },
       items: [
```
jlvandenhout, line 19: These links will not go in the footer, but rather in the bottom part of the landing page. We only should have the copyright and the socials in the footer.

jlvandenhout (changes requested):
Added a comment about the footer.

docusaurus.config.js
Dr-Electron: Are you sure about that? Because most sites contain that info in the footer, like the docusaurus docs itself

jlvandenhout: I'm taking the Figma design as a reference. There these links are contained in the last part of the main section of the landing page, including the search bar. In the docs page design they are not included and I think with good reason, as these links are already always accessible in the side bar, so having them in the footer would be a bit redundant.

Dr-Electron: Hmmm ok. I'm always for using the tools you have, but in that case we would need to implement this list ourselves. What do you think about that? Do you want to change it then? Because I don't know what the best solution would be 👍

jlvandenhout: Yeah, the landing page is not provided by docusaurus (the default one is just an example) so we have to design that ourselves anyway, so I'm happy to add them there myself.

jlvandenhout: And we need to customize the footer also anyways, to add the socials bar

Dr-Electron: I'm already on it. But that's an easy task. But I'm really not sure if we want to remove the default footer from docusaurus. As I said, the official docs also use the redundant items in the footer

jlvandenhout: I think the Docusaurus case is different, they provide community links and a secondary path to their docs not reflected in their sidebar, while in our case we would completely reflect the same structure from the sidebar in the footer (and it is also already in the navbar for that matter, although only first layer links), which imho is unnecessary duplication and I understand why it is not in the design.

Dr-Electron: They(docusaurus) also have duplicates in the sidebar. And for example if you are on mobile it's nice to have the list again at the button after you finished reading, so the probability of you reading another article is higher. I saw this redundancy on some other docs pages too, so it isn't unusual. Let's see if we get another opinion here 👍🏻

AdamCroply: Great chat. I think you seem to have figured it out. But there are a couple of points we need to think about. In general yes, the majority of links are redundant in the footer. However there are some important links included that aren't in the docs pages, such as direct links to the website, blog, etc. These would also be beneficial to have in the footer of the wiki pages also. On another note, when browsing the wiki not all menus are expanded at all times, so not all links will be visible as they are collapsed. So it may be worth keeping it in the footer for the time being, and then observing through analytics how productive the links are over time. The footer links are predominantly there as a potential quick option to the main topics in play. A guiding list if you will. No harm in them staying there for now imo :)

jlvandenhout: Ok, so let's merge this for now as discussed so everybody can play around with it and see how it feels.

jlvandenhout (approved):

PR #58, opened 2021-06-28
Implement a full search bar

jlvandenhout:
# Description of change

**NOTE: This feature is now being developed separately as a Docusaurus Plugin [here](https://github.com/jlvandenhout/docusaurus-plugin-search)**

In the Figma design we have agreed on a full search bar instead of a toggle button. This is also discussed on Discord with the UX team and is considered a better option in terms of CTA. The search bar should animate to a full screen search page.

## TODOs
- [x] Fix issue where the search page disappears when the navigation bar is hidden.
- [x] Add a button to close the search page as an alternative option to moving focus away from the search input.
- [ ] Add a section for search results and a way to fill it.
- [x] Switch back to the default navbar order with the search bar right of the dark mode button.

## Type of change


jlvandenhout:
I'm looking at using [FlexSearch](https://github.com/nextapps-de/flexsearch) together with this [hook](https://github.com/angeloashmore/react-use-flexsearch) to implement local search. This requires the index to be created during build. This can either be done by running a script during build or moving search functionality into a Docusaurus plugin and use the lifecycle apis to integrate into the Docusaurus build. The latter has my preference as it enables us to directly use the Docusauruse API.

jlvandenhout:
Also as a feature we might add [Tab to Open](https://www.chromium.org/tab-to-search)

jlvandenhout:
As per nextapps-de/flexsearch#235 I'm unable to figure out how to serialize a FlexSearch index for now, so I reverted back to using Lunr in the meantime.

jlvandenhout:
Closing in favor of #194.

PR #118, opened 2021-08-13
Added Icon support for NavBar Items

lucas-tortora:
# Description of change

Added Icon support for NavBar Items

## Type of change

- Enhancement (a non-breaking change which adds functionality)

## How the change has been tested

Changes where tested locally.  **PLEASE TEST EXTENSIVELY**

## TODO
Add Icons in EXTERNAL_DOCS_DROPDOWN_CONFIG, remove test icon from Docusaurus config

## Change checklist

- [ ] I have followed the contribution guidelines for this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas

PR #121, opened 2021-08-15
Add icons for external docs in dropdown

Dr-Electron:
# Description of change

Created a new PR to test the proposed feature of #118 and new designs for the icons


docusaurus.config.js
```diff
@@ -6,8 +6,8 @@ const katex = require('rehype-katex');
 module.exports = {
   title: 'IOTA Wiki',
   tagline: 'The complete reference for IOTA',
-  url: 'https://wiki.iota.org',
-  baseUrl: '/',
+  url: 'https://dr-electron.github.io/',
+  baseUrl: '/iota-wiki/',
```
Dr-Electron, line 10: Needs to be reverted

.github/workflows/gh-pages.yml
```diff
@@ -1,8 +1,8 @@
 name: deploy
 
 on:
-  pull_request:
-    branches: [develop]
+  push:
+    branches: [feat/add-dropdown-icons]
```
Dr-Electron, line 5: Needs to be reverted

src/theme/NavbarItem/DefaultNavbarItem.js
```diff
@@ -30,6 +33,30 @@ export function NavLink({
...
   const isExternalLink = label && href && !isInternalUrl(href);
   const isDropdownLink = activeClassName === dropdownLinkActiveClass;
+  const { isDarkTheme } = useThemeContext();
+
+  let labelContainerStyle = null;
+
+    /**
+     * If we have a labelIcon, create the element and modify the container styles.
+     */
+  let labelIcon;
```
Dr-Electron, line 43: Maybe I should set this to null to be consistent? @lucas-tortora ?

Dr-Electron (commented):
Not sure if we als already want to implement the theming ina better way...
Otherwise looks fine to me after the marked things are reverted

jlvandenhout (approved):
Just nitpicking a little, but I would either write out "Library" in full instead of "Lib" or leave it out entirely as they are part of the Libraries category. But that just something to consider. Other than that: awesome!

PR #146, opened 2021-08-20
Edit docs/learn/about-iota/energy-efficiency.md

craftingstudio:


Phyloiota:
We will discuss the links and the given numbers here with the IF research department.
Thanks for contributing. Very good write up.

Phyloiota (changes requested):
The given estimates from the pre chrysalis smartphone usage don't hold true anymore.
We will try to get a fresh comment from the researchers. But also should consider the given published studies by the IF, even if it was Raspi as a most valid source that we have atm.
The links to the external sources are also questionable, as we would aim to link only to reliable and verified sources from a domain that runs under iota.org. Will get back with input from the IF to that topic

Phyloiota:
lets keep it first out and stick with the validated numbers we have for chrysalis.
Closing for now

PR #166, opened 2021-08-25
Refactor LandingPageHeader to enable page import from docs

lucas-tortora:
# Description of change

We need to decouple the LandingPageHeader from it's text to enable the import of dev docs landing pages.  **This PR should handle only that task, once this is in place and the dev docs have also been updated we should modify the import script as well.** 

* Renamed LandingpageHeader to LandingPageHeader
* Fixed html and tags in LandingpageHeader
* Added actionList, title and tagline props to LandingpageHeader
* Changed to absolute import for components
* Removed empty tag around actionList.description
* Sending actionList, title and tagline props in src/pages/index.js

## Type of change

- Documentation Enhancment

Changes were tested locally. 

## Change checklist

- [ ] I have followed the contribution guidelines for this project


src/components/LandingPageHeader/index.js
```diff
@@ -0,0 +1,73 @@
...
+import './styles.css'
+
+
+function Action({title, link, description}) {
+    let [hovering, setHovering] = useState(false);
+    let history = useHistory();
+
+    const handleClick = (e) => {
+        e.preventDefault();
+        history.push(window.location.pathname + "/" + link);
```
jlvandenhout, line 15: Maybe it is better to use [useBaseUrl](https://docusaurus.io/docs/docusaurus-core#usebaseurl) here, because it handles edge cases like root paths and links that already contain the base path too.

```suggestion
        history.push(useBaseUrl(link));
```

jlvandenhout (changes requested):
What do you think about factoring out the action list entirely and make it configurable from the config file, just like the navbar and footer items?

lucas-tortora:
> What do you think about factoring out the action list entirely and make it configurable from the config file, just like the navbar and footer items?

Sounds good to me :) , I'll see what I can do

lucas-tortora:
Since the imported content didn't add much, I guess we can close this PR.  I'll keep the branch in my repo in case we want to revisit

PR #170, opened 2021-08-30
Reorganize custom.css

Critical94:
# Description of change

As discussed on Discord, CSS organization can be improved. I introduced a Table of Contents, added comments and put fonts to the top. The body part can still be sorted alphabetically, I have not done this. I am not sure if this is the best way to order a CSS file, but it is my personal preference.

I changed `html[data-theme='dark']` to `:root[data-theme='dark']`. 

## Type of change

- Documentation Enhancement

Tested locally

## Change checklist

- [ ] I have followed the contribution guidelines for this project

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -1,3 +1,465 @@
+/* 
+*  CSS TABLE OF CONTENTS
+*   
+*  1.0 - Font import
+*  2.0 - Root variables (Infima)
+*  2.1 - Root variables - dark theme (Infima)
+*  3.0 - Body
+*  4.0 - Logo fonts
+*/
+
+
+/*** 1.0 - Font import ***/
+  
+@font-face {
+  font-family: Inter;
+  font-style: normal;
+  font-weight: 100;
+  font-display: swap;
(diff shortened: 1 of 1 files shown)


jlvandenhout:
While I like the effort in this PR a lot has changed since and I actually want to take the ideas in this PR one step further and factor out fonts, variables and class styling into their own files, order properties within declaration blocks alphabetically and remove any hardcoded colors in favor of Infima variables. So I'm not sure what to do now with this PR, any suggestion from you guys @Critical94 or @Dr-Electron?

Critical94:
We should close this PR, your approach is even better. I think #190 aligns with this approach since we still use styles.module.css files for components whenever possible, right?

jlvandenhout:
I don't want to necessarily move away from custom.css, but I'd rather not use css modules for this project. The reason for that is we are aiming for simplicity to allow easy contribution. So if someone wants to suggest a style, I rather not have them need to look through JS code to find the correct class. I rather like co located css files with classes that follow BEM naming (adopted from Infima) to provide basic structure using Infima as much as possible and then use custom.css to override those classes when Infima does not suffice.

Critical94:
Alright got it. Easy contribution is the key factor in this project! I closed #190, and we can close this PR as well if there is no other suggestion from @Dr-Electron?

jlvandenhout:
Closing in favor of #193, which includes the ideas proposed in this PR.

PR #184, opened 2021-09-06
Simplify and center the toggle button

jlvandenhout:
Somehow the toggle icons were not centered and in any case the the component was way too complicated for what we were trying to achieve, due to remnants of what it once was (a slider). I simplified the component to only a container and the icons, which also simplifies the styling a bit. I also took the opportunity to use BEM naming for the styles.

## Type of change


src/theme/Toggle/styles.css
```diff
@@ -9,84 +9,67 @@
...
+[data-theme='dark'] .toggle .toggle__icon--unchecked,
+.toggle--checked .toggle__icon--unchecked {
   opacity: 0;
-}
-
-/*
-  Remove toggle thumb 
-*/
\ No newline at end of file
+}
```
Dr-Electron, line 75: Could you add a new line? :trollface:

Dr-Electron (changes requested):
:trollface:

Dr-Electron (approved):

PR #209, opened 2021-09-17
Edit docs/build/exchange-integration/exchange-integration-guide.md

AdamCroply:


Phyloiota:
we should not duplicate the content from the docs just into wiki pages... will be difficult to keep it up to date...

PR #210, opened 2021-09-17
Edit docs/build/getting-started/architecture.md

AdamCroply:


Phyloiota:
Hey Adam, this is all about 2.0. Should we think about having a section to build on the current Chrysalis and another section to build on the 2.0 Testnet? Otherwise, it may confuse devs that they don't find any of those topics in the current mainnet...
What do you think @Dr-Electron , @jlvandenhout ?

Dr-Electron:
> Hey Adam, this is all about 2.0. Should we think about having a section to build on the current Chrysalis and another section to build on the 2.0 Testnet? Otherwise, it may confuse devs that they don't find any of those topics in the current mainnet...
> What do you think @Dr-Electron , @jlvandenhout ?

Yeah, I thinks so. Imo we should focus everything in the wiki on 1.5 and only have (best-case) a single section for 2.0. The wiki should describe the current state in all general topics

Phyloiota:
Agree, actually... We can already think about as we will have a lot of building going on in 2.0 to divert the BUILD developer section in 1.5 and 2.0. This should help developers to decide where they want to build their apps and make live easier ....
So we should start with the architecture document here and explain the differences for developers between 1.5 Main and 2.0 Testnet. So we can provide guidance for both cases...

jlvandenhout:
I completely agree with @Dr-Electron here. 1.5 is current and stable, 2.0 is bleeding edge and under active development. Let's not overcomplicate the Wiki with unreleased stuff, but instead focus on what we can build upon now and explain it as well as we can. So I would dedicate everything to 1.5, and like @Dr-Electron suggested, have a single section that explains the future goal of 2.0.

PR #212, opened 2021-09-17
Edit docs/build/fundamentals/consensus.md

AdamCroply:


Phyloiota:
hmmm, should we wait a bit until OTV is implemented, or push this now and then change in a few weeks when OTV is life?

Dr-Electron:
> hmmm, should we wait a bit until OTV is implemented, or push this now and then change in a few weeks when OTV is life?

Just give as an ETA for OTVFPCS so that we can decide :trollface: 

I would add it now. So in the history of the wiki it got mentioned once. It's part of the IOTA history so it should be part of the Wiki history too imo.

But the real question is again: Should we move Everything IOTA 2.0 in its own section?

Issue #221, opened 2021-09-19
Start enforcing a coding standard


jlvandenhout:
I want to start using a formatter and a linter to enforce a code standard so we have a more unified codebase and it is easier for others to contribute. I'm thinking of [Prettier](https://prettier.io/) for code formatting, as it supports all the languages we use in this project out of the box, and [ESLint](https://eslint.org/) with a [React plugin](https://github.com/yannickcr/eslint-plugin-react) to lint our JavaScript. These tools have good IDE integration as well.

jlvandenhout:
@Critical94 and @Dr-Electron, any considerations or experience with alternative tools?

Critical94:
No, never used alternative tools. Prettier and eslint are the most used tools in this area. In the Prettier [installation guide](https://prettier.io/docs/en/install#eslint-and-other-linters), they even explain how to make them play nice with each other.

PR #247, opened 2021-09-24
Add properties validation

### LINKED ISSUE #245: Add property type checks for components
Currently we do not utilize the React property system in our components, while I think we should. Using [property type checks](https://reactjs.org/docs/typechecking-with-proptypes.html) prevents components rendering incorrectly if the right properties are not passed and we can provide sane defaults.

jlvandenhout:
I've added property type checking and provided the defaults where defaults were used. This will help with #221 as ESLint will complain less.

Closes #245 

## Type of change


lucas-tortora (approved):

Critical94:
There are `Failed prop type:` warnings in the console on blog page

jlvandenhout:
@Critical94 Ah thanks for letting me know, I'll fix those in a sec.

Critical94:
Did something get mixed up here? There are many unrelated changes now :monocle_face:

jlvandenhout:
> Did something get mixed up here? There are many unrelated changes now 🧐

Yes I screwed up, see Discord

jlvandenhout:
@Critical94 Could you check again? I think I solved the mess.

Critical94 (approved):

PR #277, opened 2021-09-29
Change blog post linking

Critical94:
# Description of change

This PR changes the linking of the **blog post headlines** directly to the source (medium.com, etc.). So far, they were linked to a page with the same excerpt as on the overview, which was misleading.

+ little styling fix 

Tested locally


jlvandenhout (approved):
I fixed some format issues that Prettier and ESLint reported, but other than that LGTM!

PR #288, opened 2021-10-02
Make socials configurable

jlvandenhout:
This will allow to configure the socials bar from the Docusaurus config file.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -303,6 +303,56 @@ module.exports = {
       },
       copyright: `© ${new Date().getFullYear()} IOTA Wiki. Built with Docusaurus.`,
     },
+    socials: [
+      {
+        title: 'Youtube',
+        icon: '\ue907',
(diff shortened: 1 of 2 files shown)


Dr-Electron:
The icon and title seems somehow redundant. That's why I wanted to add a LUT to the Team PR with the mapping of title to icon. Do you think it would be useful if we have a global LUT for that?

jlvandenhout:
I think that's reasonable to do, we could even check the links which social it points to, so the config just contains the links.

Dr-Electron:
> I think that's reasonable to do, we could even check the links which social it points to, so the config just contains the links.

Good point.
Do you want to implement that in this PR or should I merge and we do it with later on?

jlvandenhout:
Let's merge this and have a look at it later as we have to think about a few other things too (how to configure different icon fonts more easily for example, automatically calculate the color gradients and Discord is not easy to autodetect from the url)

Dr-Electron (approved):

PR #293, opened 2021-10-03
Remove unused pages

Dr-Electron:
# Description of change

I removed most of the unused and empty pages.
If you think one of the pages shouldn't be removed because we should write something for that topic just mention it here and I will fill it with our "help for contribution" template.

We should also have a look at the use cases section. Do we want to keep them all?

## Type of change

- Documentation Enhancement

Tested locally

## Change checklist


Critical94:
Use cases: We could remove "Private Tangle" as it is not really a use case, rather a tool to implement use cases.

Phyloiota:
I agree so far, but would keep IOTA partnerships, kinda a good topic to fill and I hav already uploaded a slide for that in my open use cases PR.
I also would keep use cases, but remove private Tangle.
The use case section can and should be filled more by the community and I can also add IF material like slides... check my PR  https://github.com/iota-community/iota-wiki/pull/294

jlvandenhout:
> I agree so far, but would keep IOTA partnerships, kinda a good topic to fill

Those partnerships are with the IF though right? They are an important topic, but not managed by the community in general. In my opinion these should be listed on the IF page and be managed there. We can link to them from the Wiki.

Any partnerships initiated from the community can be listed under use cases.

PR #303, opened 2021-10-07
Edit docs/learn/about-iota/messages.md

DafPhil:


Phyloiota:
Yes the parsing is the processing that happens in the nodes. I have made it a bit clearer in the text

PR #347, opened 2021-10-19
Change license

aleksei-korolev:
I have no lawyer license or education and whatever I say is not a legal advice.

There are few grey zones in the repository that we could clarify for contributors and users.

1. While copyright belonging to the IOTA Community sounds wholesome, IOTA Community is not a clearly defined legal entity and it is not clear who is actually allowed to make decisions to manage copyright. I suggest to go with the wiki team core members - I took the list and  preferred spelling from the Team page on the wiki, feel free to change it.
    
    When the IOTA Community will be mature enough (go DAOs!) to object this copyright claim, it will be mature enough to manage it, so you could revert this change when the time comes.

2. The MIT license governs "software and associated documentation files". This puts the contents of the `internal` folder into a grey zone: you could rule that this is either content that is a part of software, or you could rule that this is indeed "associated documentation files". CC BY 4.0 has similar spirit and ruling as MIT license, but covers any kind of works, so it would fit such content better.

This is up to you to decide if you would like to proceed with these changes. In any case, after you make this decision, there are two more grey zones we should cover.

1. In-wiki editor should have a notice that would say something like "Note that your changes will be published under IOTA Wiki or associated repositories licensing conditions."
2. We pull docu from other repositories via script to build the website, but we do not mention the license in the final product, which techically puts us in the copyright infridgement. I cannot speak for IOTA Foundation or the rest of the IOTA community, but I however doubt that there will be any lawsuit based on that :trollface:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright © Phylo, Dr. Electron, Jeroen van den Hout, Critical
+Copyright © 2021 Phylo, Dr. Electron, Jeroen van den Hout, Critical
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


aleksei-korolev:
@Phyloiota @Dr-Electron @jlvandenhout @Critical94

PR #379, opened 2021-11-02
Fix navbar positioning

jlvandenhout:
I suddenly remembered why we swizzled the navbar. We want the menu icon on the right of the navbar on mobile. To do this however, I now had to swizzle the SearchBar component as well to prevent it from obscuring the menu icon because of absolute positioning.

For correct placement of the icons in the top right, I added the icons to a dock-like div, so they are spaced correctly relative to each other and other top level navbar components.


src/theme/Navbar/styles.css
```diff
@@ -0,0 +1,7 @@
+.navbar__item--dock {
+  display: flex;
+}
+
+.navbar__item--dock > * + * {
```
Critical94, line 5: That would be a little clearer for me but the result is the same

```suggestion
.navbar__item--dock > * {
```

Critical94 (approved):
LGTM !!

src/theme/Navbar/styles.css
jlvandenhout: This breaks the toggle button for me in Chrome because it shifts the icons, but not the background. I'm going to change it to use margins btw, because it seems semantically more correct to create space between adjacent siblings.

lex111:
Hi all, as Docusaurus maintainer, I wonder why you didn't just place these buttons through absolute positioning? Then you wouldn't have swizzled components. Yes, it's not very elegant, but it's possible, and still better than the current solution with swizzling. Maybe I'm missing something, but it seems that in your case you can only use custom CSS.

jlvandenhout:
> Hi all, as Docusaurus maintainer, I wonder why you didn't just place these buttons through absolute positioning? Then you wouldn't have swizzled components. Yes, it's not very elegant, but it's possible, and still better than the current solution with swizzling. Maybe I'm missing something, but it seems that in your case you can only use custom CSS.

Thank you for visiting our repository! The reason for me is that it feels more natural to place items correctly in the structure and let the layout do its thing, rather then having to calculate positions ourselves in CSS.

lex111:
I understand you, it's just that having swizzled components has a number of disadvantages, for example you lose future changes/improvements in these components. Therefore, if possible, it is better to try to achieve changes in the layout through custom CSS.

PR #391, opened 2021-11-17
Edit internal/learn/about-iota/data-transfer.md

luckybusted:
Added a missing link to the IOTA 2.0 Documentation on the page: https://wiki.iota.org/learn/about-iota/data-transfer


jlvandenhout (approved):
Thanks for adding the link!

PR #403, opened 2021-11-30
Feat/fullscreen images

jlvandenhout:
We want full screen image functionality. Currently we just wrap images in a link to the image source, to open it in a separate tab in the browser. This feature implements an alternative approach where we provide a top level image container and wrap images in the markdown files in an image component that opens the top level image container with the source image when clicked.

The upside to this is that we can use plain markdown as source without the arbitrary linking and handle any image interaction on the React side where it belongs.


Critical94 (approved):
Very cool, much better than before. Nothing to improve from my side.

PR #482, opened 2022-03-10
Remove unused assets

jlvandenhout:
# Description of change

I searched far and wide, but I don't think we use these anymore anywhere.


Dr-Electron (approved):
Yeah, we don't 😅 
This was way to long ago 😆

Critical94 (approved):

PR #512, opened 2022-04-22
Learn - Value Transfer Update

charlesthompson3:
# Description of change

Content in the Learn: Value Transfer section has been updated in lieu of the Stardust updates.

## Type of change

- Documentation Fix
- Documentation Enhancement

## Change checklist

### DIFF
--- a/internal/learn/about-iota/value-transfer.md
+++ b/internal/learn/about-iota/value-transfer.md
@@ -6,13 +6,13 @@ description: Learn how the IOTA Cryptocurrency Protocol transfers the native IOT
 
 # IOTA Value Transactions
 
-This page explains how IOTA tokens are sent and received in the IOTA network and links to more detailed content.
+This page explains how IOTA tokens are sent and received in the IOTA network with references to more detailed content.
 
-## Simple example:
+## Simple Example
 
 To explain how IOTA tokens are transferred from one address to another, this topic follows three characters: Alice, Bob, and Charlie.
 
-### The goal
+### The Goal
 
 Charlie wants to transfer 10i to Bob.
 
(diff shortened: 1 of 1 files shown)


internal/learn/about-iota/value-transfer.md
```diff
@@ -6,82 +6,82 @@ description: Learn how the IOTA Cryptocurrency Protocol transfers the native IOT
...
+### The Goal
 
 Charlie wants to transfer 10i to Bob.
 
 ### What is in the Tangle
 
 In this example, Alice had 10i that she transferred to Bob. Therefore, the Tangle now includes a value message that transferred those 10i to Bob.
 
-### Choosing where to attach messages
+### Choosing Where to Attach Messages
```
lzpap, line 23: I don't get the `In this example, Alice had 10i that she transferred to Bob` part. Where does Alice ocme into play when the goal is that Charlie tranfers 10i to Bob?

lzpap: So what is confusing is that:
 - We state that Alice transferred 10i to Bob.
 - The example is about how Charlie transfers 10i to Bob.
 - Why the hell do we need Alice at all then?

Let's get rid of this Alice transferred bob thingy. And later, when you make the double spend example just say: but what if Charlie submitted two trasanctions, one that transfers Bob 10i and one that transfers Alice 10i?

jlvandenhout (approved):

Dr-Electron:
Why is this a stardust doc?

charlesthompson3:
> Why is this a stardust doc?

Small node operation update/correction, I believe.

jlvandenhout:
Both @Dr-Electron and I think we can just target the `main` branch here, because the changes are more editorial than Stardust related changes, if that's okay with you @charlesthompson3?

charlesthompson3:
> Both @Dr-Electron and I think we can just target the `main` branch here, because the changes are more editorial than Stardust related changes, if that's okay with you @charlesthompson3?

Branch switched and ready for approval.

Dr-Electron (approved):

PR #513, opened 2022-04-22
Learn - IOTA Messages Updates

charlesthompson3:
# Description of change

Content in the Learn: Value Transfer section has been updated in lieu of the Stardust updates which includes:

- updating the utxo model and payloads
- removing "payloads" entirely
- Adding reference links where applicable

## Type of change

- Documentation Fix
- Documentation Enhancement

## Change checklist

### DIFF
--- a/internal/learn/about-iota/messages.md
+++ b/internal/learn/about-iota/messages.md
@@ -8,7 +8,7 @@ description: This topic explores messages, models, and payloads that encompass t
 
 A message is an object that nodes gossip around in the network. It always references one to eight other messages, which are known as **parents**. It is stored as a vertex on the Tangle data structure maintained by the nodes.
 
-Messages can contain payloads. Some of them are core payloads that are processed by all nodes as part of the core protocol. Others are community payloads that enable the building of new functionalities on top of the Tangle. And some payloads have other nested payloads embedded inside. So, the parsing of the message is done layer by layer, to ensure embedded payloads also have a correct syntax structure.
+Messages can contain payloads. Some of them are core payloads that are processed by all nodes as part of the core protocol. Others are community payloads that build up new functionalities on top of the Tangle. And some payloads have other nested payloads embedded inside. So, the parsing of the message is done layer by layer, to ensure embedded payloads also have a correct syntax structure.
 
 ## UTXO
 
@@ -18,7 +18,7 @@ The UTXO model defines a ledger state where balances are not directly associated
 
 ![utxo-model](/img/learn/about-iota/utxo.png)
 
-So, the UTXO is a part of a larger, self-contained, and flexible message structure known as a **payload**. This approach is meant to enable a self-contained message structure defining the data of the entire transfer as a payload to be embedded into a message.
+So, the UTXO is a part of a larger, self-contained, and flexible message structure known as a **payload**. This approach is helps enable a self-contained message structure defining the data of the entire transfer as a payload to be embedded into a message.
 
 ## Message Payloads
 
@@ -29,7 +29,7 @@ A transaction payload (defined in [TIP-0020](https://github.com/lzpap/tips/blob/
 1. The Transaction Essence part, which contains the inputs, outputs, and an optional embedded payload.
 2. The Unlock Blocks, which unlocks the Transaction Essence's inputs. In case the unlock block contains a signature, it signs the entire Transaction Essence part.
 
-In general, all parts of a transaction payload begin with a byte describing the type of the given part to keep the flexibility to introduce new types or versions of the given part in the future.
+In general, all parts of a transaction payload begin with a byte describing the type of a given part. This helps keep the flexibility of introducing new types or versions of the given part in the future.
 
 And, as mentioned above, the payload part of a Transaction Essence can hold an optional payload. This payload does not affect the validity of the Transaction Essence. If the transaction is not valid, then the payload must also be discarded.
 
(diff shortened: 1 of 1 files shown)


internal/learn/about-iota/messages.md
```diff
@@ -1,68 +1,53 @@
...
 
 ## UTXO
 
-Previously, the IOTA protocol used transactions (which were vertices in the Tangle), where each transaction defined either an input or an output. A grouping of those input/output transaction vertices made up a bundle that transferred the given values as an atomic unit. But this approach was seen as too time-consuming. So, we adopted a new transaction structure called the **unspent transaction outputs **(UTXO).
+Previously, the IOTA protocol used transactions (which were vertices in the Tangle), where each transaction defined either an input or an output. A grouping of those input and output transaction vertices made up a bundle that transferred the given values as an atomic unit. But this approach was seen as too time-consuming. So, we adopted a new transaction structure called the **unspent transaction outputs **(UTXO)**.
```
jlvandenhout, line 15:
```suggestion
Previously, the IOTA protocol used transactions (which were vertices in the Tangle), where each transaction defined either an input or an output. A grouping of those input and output transaction vertices made up a bundle that transferred the given values as an atomic unit. But this approach was seen as too time-consuming. So, we adopted a new transaction structure called the **unspent transaction outputs (UTXO)**.
```

internal/learn/about-iota/messages.md
```diff
@@ -1,68 +1,53 @@
...
 
 The UTXO model defines a ledger state where balances are not directly associated with addresses but with the outputs of transactions. In this model, transactions specify the outputs of previous transactions as inputs, which are consumed to create new outputs. A transaction must consume the entirety of the specified inputs.
 
 ![utxo-model](/img/learn/about-iota/utxo.png)
 
-So, the UTXO is a part of a larger, self-contained, and flexible message structure known as a **payload**. This approach is meant to enable a self-contained message structure defining the data of the entire transfer as a payload to be embedded into a message.
+So, the UTXO is a part of a larger, self-contained, and flexible message structure known as a **payload**. This approach is helps enable a self-contained message structure defining the data of the entire transfer as a payload to be embedded into a message.
```
jlvandenhout, line 21:
```suggestion
So, the UTXO is a part of a larger, self-contained, and flexible message structure known as a **payload**. This approach helps to enable a self-contained message structure defining the data of the entire transfer as a payload to be embedded into a message.
```

internal/learn/about-iota/messages.md
```diff
@@ -1,68 +1,53 @@
...
-### Transaction payload
-
-A transaction payload is made up of two parts:
+A transaction payload (defined in [TIP-0020](https://github.com/lzpap/tips/blob/tx-updates/tips/TIP-0020/tip-0020.md)) is made up of two parts:
 
 1. The Transaction Essence part, which contains the inputs, outputs, and an optional embedded payload.
 2. The Unlock Blocks, which unlocks the Transaction Essence's inputs. In case the unlock block contains a signature, it signs the entire Transaction Essence part.
 
-In general, all parts of a transaction payload begin with a byte describing the type of the given part to keep the flexibility to introduce new types/versions of the given part in the future.
+In general, all parts of a transaction payload begin with a byte describing the type of a given part. This helps keep the flexibility of introducing new types or versions of the given part in the future.
```
jlvandenhout, line 32:
```suggestion
In general, all parts of a transaction payload begin with a byte describing the type of a given part. This allows the introduction of new types or versions of the given part in the future.
```

jlvandenhout (changes requested):

lzpap (approved):

jlvandenhout (approved):

Dr-Electron:
So as this already explains stardust stuff, how do we merge it? Once stardust is on mainnet? Because otherwise we need to separate the content as long as we have two concepts

charlesthompson3:
> So as this already explains stardust stuff, how do we merge it? Once stardust is on mainnet? Because otherwise we need to separate the content as long as we have two concepts

Yes, the former I think.

jlvandenhout:
We've internally discussed the proposed changes in the Stardust PRs and we came to the conclusion they are not quite in line with what we had in mind for the Stardust related changes to the Wiki.

Contrary to the build section documentation, where we necessarily will have versioning to differentiate between networks and releases of software, it has been decided that we will have a single version for the learn and participate sections. The reason is that the documentation of these sections should be at such an abstraction level that the implementation details are preferably non-existent and the differences are relatively small, so that it should suffice to just point out the differences that do exist between networks where necessary within the content. We keep the state of mainnet as the source of truth for the main content and indicate where Shimmer, and later maybe Coordicide, deviate from mainnet.

To make these deviations clear to the reader and to the maintainers of the content, we have the following at our disposal:
1. [Admonitions](https://docusaurus.io/docs/markdown-features/admonitions): probably best to indicate differences in concepts and explanations that are already implemented or will be in the near future.
2. [Tabs](https://docusaurus.io/docs/markdown-features/tabs): probably best to indicate differences in tutorials and guides where necessary, like tables of message types, commands or example code.
3. [The `The future of IOTA` section](https://wiki.iota.org/learn/future/future-of-iota): probably best to indicate planned differences that will be implemented at a later stage or are still up for discussion.

We are aware that this requires changes to the current PRs, but that should mostly involve adding the proposed updates to a note admonition instead of changing the existing text itself. Only the updates that involve planned differences or differences under discussion should be moved to the `The future of IOTA` section.

PR #517, opened 2022-04-23
Fix mega menu

jlvandenhout:
# Description of change

I removed some unnecessary swizzled components and I re-implemented the megamenu based on the new dropdown component provided by Docusaurus in beta.18. To make it work, I needed to swizzle the primary menu component of the mobile navbar, so it properly passes the mobile property to the navbar item components.

This is a temporary fix, as we will be able to add custom components properly soon.

## Type of change

Tested locally and also by adding a versioning dropdown, to make sure that doesn't break.

## Change checklist


Dr-Electron:
LGTM. Although we should still fix the Safari Layout. Something is off there 😅

Dr-Electron:
To me it looks like the arrow component defined with [`::after`](https://github.com/iota-community/iota-wiki/blob/b020ce8428caf56625d6543d311672ec353d71d9/src/css/custom.css#L6) takes to much space and that's why together with `word-wrap: break-word;` it "breaks". So I think we need to restrict the size of the material icon. And is it intentional that the arrow icon doesn't show? Because if not we need to fix that too 😅 

Bildschirmfoto 2022-04-25 um 13 52 10

Dr-Electron (approved):

huhn511:
Works well on Safari now!

huhn511 (approved):
LGMT 👍

PR #520, opened 2022-04-26
added wasp node to test

anistark:
# Description of change

Adding a tutorial for Running a wasp node.

## Type of change

- Documentation Enhancement

Make sure to provide instructions for the maintainer as well as any relevant configurations.

## Change checklist


content/tutorials/content/single-page-tutorials/docusaurus.config.js
```diff
@@ -13,6 +13,13 @@ module.exports = {
         route: 'send-iota-tokens-with-javascript',
         tags: ['text'],
       },
+      {
+        title: 'Run a Wasp Node',
+        description: 'In this tutorial you will learn how to run a wasp node.',
+        preview: 'IOTA-Smart-Contract-Tutorials-A.jpg',
+        route: 'https://wiki.iota.org/smart-contracts/guide/chains_and_nodes/running-a-node',
```
Dr-Electron, line 20:
```suggestion
        route: 'smart-contracts/guide/chains_and_nodes/running-a-node',
```

Please don't use a domain. Only the path is enough. Otherwise test build would break 👍

Dr-Electron:
Why are there so many images? 😅

Dr-Electron:
Also not sure if we want to add this to the tutorials section. Or maybe there is a better way to do this. As the TW tagged all docs. So they should already have the tutorial tag. @jlvandenhout what do you think?

jlvandenhout:
We had a call about this and decided that there was no harm in adding it already, so people who search the tutorial page can find it. When the TWs actually come to moving the content, it is only a matter of updating the route in this config.

Dr-Electron:
> We had a call about this and decided that there was no harm in adding it already, so people who search the tutorial page can find it. When the TWs actually come to moving the content, it is only a matter of updating the route in this config.

I was more talking about maybe adding some logic to add all stuff tagged as tutorial?
Or is the plane to move everything into the tutorial section? Because not sure if that's always possible 🤔

jlvandenhout:
> I was more talking about maybe adding some logic to add all stuff tagged as tutorial?
> Or is the plane to move everything into the tutorial section? Because not sure if that's always possible 🤔 

I see, didn't think of that. I guess you are right it makes more sense to keep the tutorial configs colocated to the docs  they point to. So when docs are tagged as tutorial, we add the tutorial config to the related `docusaurus.config.js`. This would also make more sense when testing locally.

anistark:
> > I was more talking about maybe adding some logic to add all stuff tagged as tutorial?
> > Or is the plane to move everything into the tutorial section? Because not sure if that's always possible 🤔
> 
> I see, didn't think of that. I guess you are right it makes more sense to keep the tutorial configs colocated to the docs they point to. So when docs are tagged as tutorial, we add the tutorial config to the related `docusaurus.config.js`. This would also make more sense when testing locally.

We would want to reduce duplicating content wherever possible.
Cause every small change would lead to a cascade of unnecessary changes along with otherwise.
To it, Either if you figure out the tagging of docs pages or putting here manually. Both works.

Also, I thought of adding all images at once and then if this format works, can add all different object blocks in respective PRs.

jlvandenhout:
No duplication is needed. The tutorial page simply gathers all configured tutorial plugins, it doesn't actually matter where the tutorials are configured or what they point to. So I'd say in terms of modularity it is actually preferable to have the tutorial plugin configuration right in the `docusaurus.config.js` that is associated with the documentation it points to, so that would be the Wasp documentation config in this case.

anistark:
> No duplication is needed. The tutorial page simply gathers all configured tutorial plugins, it doesn't actually matter where the tutorials are configured or what they point to. So I'd say in terms of modularity it is actually preferable to have the tutorial plugin configuration right in the `docusaurus.config.js` that is associated with the documentation it points to, so that would be the Wasp documentation config in this case.

Could you share an example of it?

Dr-Electron:
> you share an example of it?

It's basically what you did, but just in the wasp repo directly. By adding
```
'@iota-wiki/plugin-tutorial',
      {
        title: 'Run a Wasp Node',
        description: 'In this tutorial you will learn how to run a wasp node.',
        preview: 'IOTA-Smart-Contract-Tutorials-A.jpg',
        route: 'smart-contracts/guide/chains_and_nodes/running-a-node',
        tags: ['text', 'video'],
      },
```

there. Just on mobile, so not 100% sure if i copied the right thing 😅
But this will only work once we merged the new build system PR and they do it too

jlvandenhout:
Tutorial cards are configured by a separate tutorial plugin config for each one, so I updated the config to reflect that. (I also fixed a small dependency issue that I should have fixed earlier, but now at least this branch builds without checking out the other tutorial submodules).

Dr-Electron (approved):

PR #526, opened 2022-04-30
Add new issue forms

Dr-Electron:
# Description of change

Use new issue form to make reporting bugs, requesting features and topics more precise and easy to fit in infos.

# Open Questions

- [ ] Do we need more options, or is bug report, feature and topic request enough?
- [ ] Do we want to add additional info in the issue interface? Like for example the discord link [here](https://github.com/iotaledger/template/issues/new/choose)
- [ ] Which labels should we use?


jlvandenhout:
I think adding an option to create an issue specifically to content would make sense. So we have bug reports and feature requests for Wiki functionality and documentation issues and topic requests for content? Didn't think of a good name for that yet though.

jlvandenhout:
For labels we then would have: `bug`, `feature`, `topic` and `content` as category labels.

I was never fond of priority labels, but I'm ready to be convinced. Also I don't see the point in having status labels like `won't fix` or `duplicate`, because that would be something we comment as reason for closing the issue anyway, but again I'm open for discussion there.

Labels like `needs information` and `confirmed`/`accepted` I think would be helpful, as those are all status while the issue is still open.

Dr-Electron:
> For labels we then would have: `bug`, `feature`, `topic` and `content` as category labels.
> 
> I was never fond of priority labels, but I'm ready to be convinced. Also I don't see the point in having status labels like `won't fix` or `duplicate`, because that would be something we comment as reason for closing the issue anyway, but again I'm open for discussion there.
> 
> Labels like `needs information` and `confirmed`/`accepted` I think would be helpful, as those are all status while the issue is still open.

Yeah we can work on that. Those are mostly the default ones from Github. I asked more for the issue templates.
But for sure we also need to clean up our labels 😅

README.md
```diff
@@ -23,9 +23,9 @@
     <a href="https://wiki.iota.org"><strong>Explore the Wiki »</strong></a>
```
lucas-tortora, line 23: Just out of curiosity, why is the readme using html instead of md?

lucas-tortora (approved):

PR #541, opened 2022-05-12
Fix smaller bugs in new build system

Dr-Electron:
# Description of change

I fixed the regex for edit url so that it works for new and old builds. We can remove the external regex once we updated all docs

Part of #481 

## Type of change

Tested locally

## Change checklist

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -21,7 +21,7 @@ jobs:
     if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     env:
-      BUILD: '*'
+      BUILD: '**'
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -21,7 +21,7 @@ jobs:
     if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     env:
-      BUILD: '**'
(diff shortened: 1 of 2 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     env:
-      BUILD: '*'
+      BUILD: '**'
```
jlvandenhout, line 24: I'm not sure if this is the way to go. This will import all descendants that have a config, including descendants that are already included in the config of their parent. For example the tutorials are gathered in a parent tutorial config, here this pattern would cause duplicate imports for all tutorials right?

jlvandenhout (changes requested):

.github/workflows/gh-pages.yml
jlvandenhout: I mean for now it doesn't matter, since there is nothing yet in the build directory, but still.

Dr-Electron: I think I want the same structure for versioned and unversioned docs. So I will just go for `*/*/documentation`

Dr-Electron:
```suggestion
      BUILD: '*/*/documentation'
```

jlvandenhout (approved):

Critical94 (approved):

PR #552, opened 2022-05-25
Checkout submodules on test build too

jlvandenhout:
# Description of change

Currently the test build fails because we are in the process of switching to submodules for external repos, which when not checked out result in a lock file mutation when installing dependencies.

## Type of change

The scripts are tested locally. The workflows should run in this PR and when merged.

## Change checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,6 @@
 name: test
 
 on:
-  workflow_dispatch:
   pull_request:
     branches: [main]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,31 @@
+name: test
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    env:
+      BUILD: '*/*/documentation'
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Needed for showLastUpdateTime to work
+      - name: Init and update submodules
+        run: |
+          yarn checkout:remote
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
+name: test
+
+on:
+  workflow_dispatch:
```
Dr-Electron, line 4: Do we need a workflow_dispatch here? I mean it doesn't hurt to have it. Just asking if we want a clean setup or just keep it

package.json
```diff
@@ -15,7 +15,9 @@
     "lint": "eslint --cache --fix src",
     "format": "prettier --write .",
     "prepare": "husky install",
-    "cli": "iota-wiki-cli"
+    "cli": "iota-wiki-cli",
+    "checkout": "git submodule update --init --recursive",
```
Dr-Electron, line 19: I though this doesn't work? Why do we keep it?

Dr-Electron (changes requested):

package.json
jlvandenhout: For local development it's still useful.

.github/workflows/test.yml
jlvandenhout: Yeah you are right, we can remove this.

Dr-Electron:
```suggestion
```

Dr-Electron: I leave the honour of removing it to you :trollface:

package.json
Dr-Electron: Hmm. Ok. I would include it in the cli tho. But we can leave it for now 👍

kilianhln (approved):

PR #554, opened 2022-05-26
Edit internal/learn/about-iota/value-transfer.md

sexypotato-wagmi:


Dr-Electron:
Thanks for mentioning it. Fixed with e097076 👍

PR #560, opened 2022-05-27
Fix footer list styling

jlvandenhout:
# Description of change

Docusaurus redid the footer, causing the list items to be styled incorrectly in our build. This updates the footer styling to the new component.

## Type of change

Tested with a local Wiki build.

## Change checklist

### DIFF
--- a/src/theme/Footer/index.tsx
+++ b/src/theme/Footer/index.tsx
@@ -1,146 +1,35 @@
-/**
- * Copyright (c) Facebook, Inc. and its affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- */
+import React from 'react';
+
+import {useThemeConfig} from '@docusaurus/theme-common';
+import FooterLinks from '@theme/Footer/Links';
+import FooterLogo from '@theme/Footer/Logo';
+import FooterCopyright from '@theme/Footer/Copyright';
+import FooterLayout from '@theme/Footer/Layout';
+import Social from '@site/src/components/Social';
+
 
 /**
  * REASONS TO SWIZZLE:
  * - The socials bar needed full width, so needed to be placed outside the default container.
  */
 
-import React from 'react';
-import clsx from 'clsx';
-import Link from '@docusaurus/Link';
-import { FooterLinkItem, useThemeConfig } from '@docusaurus/theme-common';
-import useBaseUrl from '@docusaurus/useBaseUrl';
-import ThemedImage, { Props as ThemedImageProps } from '@theme/ThemedImage';
-import Social from '@site/src/components/Social';
-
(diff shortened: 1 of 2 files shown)


src/theme/Footer/index.tsx
```diff
@@ -1,146 +1,35 @@
...
+import FooterLinks from '@theme/Footer/Links';
+import FooterLogo from '@theme/Footer/Logo';
+import FooterCopyright from '@theme/Footer/Copyright';
+import FooterLayout from '@theme/Footer/Layout';
+import Social from '@site/src/components/Social';
+
 
 /**
  * REASONS TO SWIZZLE:
  * - The socials bar needed full width, so needed to be placed outside the default container.
```
Dr-Electron, line 13: Not sure if we talked about this, but would it make sense to mention the version here from which this was swizzled?

Dr-Electron (changes requested):

src/theme/Footer/index.tsx
jlvandenhout: We talked about it, but I got confused how this would help. Only way I see that working is we create a commit with the swizzled component, before making any changes right? So you can mark that commit, then apply any changes and you can always compare with the commit of the fresh swizzle.

Dr-Electron: Couldn't you compare the referenced version with your swizzled file. Wouldn't that already help? 🤔

Dr-Electron: Because the referenced version is the unchanged file, right?

Dr-Electron: Lets say you swizzled this component from `2.0.0-beta.20`. Which you would add to the comment. Now I can look for tag `2.0.0-beta.20` and search the footer component: https://github.com/facebook/docusaurus/blob/v2.0.0-beta.20/packages/docusaurus-theme-classic/src/theme/Footer/index.tsx

jlvandenhout: But would that allow to diff between your file and the original Docusaurus file? The nice thing about marking the commit containing the freshly swizzled component is editors can show the differences between the file version of that commit and the current version.

Dr-Electron: That's why diff plugins or options in editors exist :trollface:

jlvandenhout: Yes but I wasn't sure if that also works across repos. But if you say that works, then we can just mention the Docusaurus version yes.

jlvandenhout: Ok, since we can't rely on individual commits existing in history, we are going with Docusaurus versioning.

Dr-Electron (approved):

kilianhln (approved):

PR #598, opened 2022-07-04
Add tutorials and simplify content structure

Dr-Electron:
# Description of change

I simplified the content folder structure and added the first tutorials

## Type of change

- Documentation Enhancement

## Change checklist


.github/workflows/deploy.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     env:
-      BUILD: '*/*/documentation,*/*'
+      BUILD: '*/*/*/documentation,*/*/*,*/*'
```
jlvandenhout, line 12:
```suggestion
      CONTENT: '*/*/*/documentation,*/*/*,*/*'
```

.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     env:
-      BUILD: '*/*/documentation'
+      BUILD: '*/*/*/documentation,*/*/*,*/*'
```
jlvandenhout, line 11:
```suggestion
      CONTENT: '*/*/*/documentation,*/*/*,*/*'
```

content/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": {
+        "require-glob": "^4.1.0"
+    }
+}
```
jlvandenhout, line 5:
```suggestion
{
    "dependencies": {
        "require-glob": "^4.1.0"
    },
    "workspaces": {
        "build",
        "tutorials"
    }
}
```

I'd like to suggest to define the workspaces here to keep the "tree structure" consistent. See corresponding update to root level `package.json`.

package.json
```diff
@@ -83,6 +83,7 @@
   },
   "packageManager": "yarn@3.2.0",
   "workspaces": [
+    "content",
     "content/build",
     "content/tutorials"
```
jlvandenhout, line 88:
```suggestion
    "content"
```
 See earlier note

jlvandenhout (changes requested):

content/package.json
jlvandenhout: Actually now I think of it we can remove the `package.json` files in build and tutorials folders and add `*/*/*/documentation` and `*/*/documentation` here.

Dr-Electron: Actually we can't. I thought about it. But maybe I'm overseeing something. But we also have none `documentation` workspaces and a different level of depth in `tutorials`. Am I missing something?

jlvandenhout: In case of non-`documentation` workspaces we need to add the workspace explicitly anyways, because we can't do `*/*` or similar if some of the projects are nodejs projects themselves.

jlvandenhout: Which is another reason why it might be worth another look at how we utilize git to checkout projects, because if we can sparse checkout only documentation directories, that problem would go away and our logic would reduce to a single recursive glob. I think latest changes of git actually allow this, but I have to check if GitHub implemented that yet.

PR #640, opened 2022-07-18
Update sidebars.ts

Dr-Electron:


jlvandenhout:
Would it make sense to move Wasp to Nodes if we rename it like this?

Dr-Electron:
No. Wasp is not a Node :trollface: 

But to be honest I would keep it under Extensions. I mean we would move Wasp, but not Chronicle? And in the end they are all "just" extensions of L1 nodes. But we can also gather further feedback from the TW or someone else if you want

jlvandenhout:
Yeah I guess you could go either way with both of them. Ok let's leave it like this and see what we get for feedback 👍

jlvandenhout (approved):

lucas-tortora (approved):

PR #687, opened 2022-08-03
Feat - Remove Libraries Section

lucas-tortora:
# Description of change

Until we have some content for the libraries section:

* Removed the libraries section as it's the same content as in the home page.  
* Added # link to the home page until we have proper libraries content.

## Type of change

Wiki was built locally

## Change checklist


docusaurus.config.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
         },
         {
           label: 'Libraries',
-          to: 'categories/libraries',
+          to: '/#libraries',
```
aleksei-korolev, line 95: Out of curiosity, what does that change do?

Dr-Electron: It scrolls to `Libraries` on the landing page

aleksei-korolev: right :facepalm:

aleksei-korolev (approved):

Dr-Electron (approved):
LGTM. Although I can't remember why we decided against just linking to the landing page. But I think this is fine 😅

jlvandenhout:
> LGTM. Although I can't remember why we decided against just linking to the landing page. But I think this is fine 😅

Because the landing page is just a showcase and not necessarily a complete reference to everything. So I wanted to have this section already as a separate page, so we can improve upon it and have a place to write content to @lucas-tortora . But if we want content first, before creating the section, that's fine by me.

PR #695, opened 2022-08-11
Update value-transfer.md

h0rs3fa11:
# Description of change
```
Both of Alice's messages could not be part of the ledger state because it would result in a negative balance for Alice’s address: minus 10i.
```
According to the context, it should to be Charlies's message and Charlie's balance

## Type of change

- Documentation Fix


Dr-Electron (approved):
Thanks for the PR

jlvandenhout (approved):

PR #718, opened 2022-08-30
Update data-transfer.md

cbries:
The document mixes the required number of parent IDs, at the beginning it is `2 - 8`, later it is stated `1 - 8` two times.


Dr-Electron (approved):
Thanks for the fix ❤️

PR #719, opened 2022-08-30
Update value-transfer.md

cbries:
some cross references to make the documentation more clear

### DIFF
--- a/internal/learn/about-iota/value-transfer.md
+++ b/internal/learn/about-iota/value-transfer.md
@@ -27,7 +27,7 @@ To transfer 10i to Bob, Charlie must attach a value message to the Tangle. For t
 
 To get these tip messages, Charlie requests them from a node.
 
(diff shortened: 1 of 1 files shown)


internal/learn/about-iota/value-transfer.md
```diff
@@ -27,7 +27,7 @@ To transfer 10i to Bob, Charlie must attach a value message to the Tangle. For t
 
 To get these tip messages, Charlie requests them from a node.
 
-The node selects these messages by using an algorithm called **Uniform Random Tip Selection** (URTS). This algorithm selects between one and eight valid tip messages that lead to a valid ledger state.
+The node selects these messages by using an algorithm called **Uniform Random Tip Selection** [URTS](https://wiki.iota.org/IOTA-2.0-Research-Specifications/4.3TipSelectionAlgorithm). This algorithm selects between one and eight valid tip messages that lead to a valid ledger state.
```
Dr-Electron, line 30:
```suggestion
The node selects these messages by using an algorithm called **Uniform Random Tip Selection** [URTS](https://iotaledger.github.io/tips/tips/TIP-0003/tip-0003.html). This algorithm selects between one and eight valid tip messages that lead to a valid ledger state.
```

As it is IOTA we should reference tips. As the current TS and future TS will be different

Dr-Electron (approved):

PR #724, opened 2022-08-31
Improve Tangle documentation

samuel-rufi:
# Description of change

Added some improvements on top of the existing Tangle documentation.

### DIFF
--- a/internal/learn/about-iota/tangle.md
+++ b/internal/learn/about-iota/tangle.md
@@ -28,18 +28,17 @@ This section refers to the most recent "Tangle" research findings:
 A blockchain consists of a growing list of records also called blocks that are linked together using cryptography.
 Transactions can only become part of the ledger if they are included in a newly issued block.
 In the case that not all new transactions fit into one block, some of them have to be postponed to the next block or later.
+Accelerating block creation and/or increasing block sizes is not a solution, as it is well known that this would compromise safety. In other words, to guarantee the security of the system, the throughput of the system is artificially suppressed so that each block propagates fully before the next block is created.
 
 ![An animation of a blockchain where new transactions have to pass through a narrow gap one by one.](/img/learn/blockchain-bottleneck.gif 'Click to see the full-sized image.')
 
 _The blockchain bottleneck._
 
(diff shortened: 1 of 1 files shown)


internal/learn/about-iota/tangle.md
```diff
@@ -1,42 +1,49 @@
...
 image: /img/iota-wiki.png
 keywords:
-  - explanation
+- explanation
 ---
 
 # The Tangle
 
-The Tangle is a structure that holds all necessary information to keep track of token ownership. It is a directed acyclic graph (DAG) of messages, where each newer message is attached to two to eight older ones. The protocol can process these various attachments in parallel.
+The Tangle is IOTA's distributed ledger. It's a data structure replicated across a network of computers (also called 'nodes') that contains all the information necessary to track token ownership. It forms a directed acyclic graph of blocks (a block-DAG), where each newer block is attached to multiple older ones.
```
Dr-Electron, line 12: So about blocks... Not sure if we should already use that in the IOTA part of the wiki as this is something coming with Shimmer. So it isn't "live" yet on IOTA 🤔

samuel-rufi: Good point. I think calling it further "Message" doesn't help people to understand our tech. And the wiki should be all about understanding, so we shouldn´t wait for Stardust to be finally deployed to the IOTA mainnet.
Also we are talking here about a general Tangle documentation, it should be used for IOTA and Shimmer.

Specific documentation about the "Message" object in IOTA should stay unchanged IMO.

lucas-tortora: Seeing that we'll be separating the IOTA and Stardust wikis, it shouldn't be too much work since most of the stardust docs already use blocks instead of messages. @Dr-Electron, would it be possible to have a `stardust` branch in the IOTA wiki for now so we can isolate this and keep `messages` in the `main` branch and `blocks` in the `stardust` branch?

Dr-Electron: 😅 We are probably going down the `separate folder` way. So basically like docusaurus versioning works.
But Sam thinks we could just use blocks in the high level descriptions of IOTA too 🤔

Dr-Electron: So we agree. And can merge. Makes sense 😆

Dr-Electron (approved):

PR #741, opened 2022-09-14
intro

aleksei-korolev:
shimmie


Dr-Electron:
:wen: tho? :trollface:

aleksei-korolev:
@lucas-tortora could you please take a look?

Dr-Electron:
Looks good to me. But why do we remove everything from the future section? Can't we reuse some stuff now that we have a shimmer wiki? Explain SCs, storage deposit (community article) and so on 🤔

Dr-Electron:
Ahh and the landing page and navbar need to be update. But we can just merge and do it then

samuel-rufi (changes requested):

shimmer/learn/welcome.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Basics
+
+### Nodes
+
+Shimmer is a network made out of *nodes*, computers that run the Shimmer node software. Anyone can join and leave the network, and nodes are run by individual parties. If you own an average computer, you can run your own Shimmer node, too.
+
+### Blocks and Outputs
+
+The structure that stores the data is called the [Tangle](./tangle.md). It consists of separate *blocks*, each of which stores some small data change. By including blocks into the Tangle, nodes agree on how the Tangle looks like. To get included, each block should reference from two to eight older blocks, and be referenced by a newer block. It means that there is always a front of the most recent blocks: you even can [see](https://explorer.shimmer.network/testnet/visualizer/) it. The more new pending blocks there are, the wider the front and the more points are there for newer blocks to attach. This scales really well.
```
samuel-rufi, line 24:
```suggestion
The structure that stores the data is called the [Tangle](./tangle.md). It consists of records, called blocks, each of which stores some small data change. The blocks form a directed acyclic graph (a block-DAG), where each newer block is attached to multiple older ones. It means that there is always a front of the most recent blocks: you can [see](https://explorer.shimmer.network/testnet/visualizer/) it. The wider the front of recent blocks the more points are there for new blocks to attach.
```

samuel-rufi (changes requested):

shimmer/learn/welcome.md
```diff
@@ -0,0 +1,50 @@
...
+
+Now, when the nodes agreed on the Tangle state, they can extract useful information from it: for example, who owns what tokens. Each block specifies outputs. In the most basic form, outputs specify an amount of tokens and their owner. Blocks exhaust existing outputs and create new outputs with the same total amount of tokens, but with different distribution and owners. Each output can be spent only once: all future blocks that try to exhaust it will be rejected by the nodes. You can ask a node to go through all blocks for a list of created and exhausted outputs that mention you as an owner. If you strike out all exhausted outputs from the list of created outputs, you will get a list of *unspent outputs*. The sum of their tokens will make your token balance.
```
samuel-rufi, line 26:
```suggestion
Now, when the nodes agreed on the Tangle state, they can extract useful information from it: for example, who owns what tokens. According to the UTXO model, blocks specify outputs. In the most basic form, outputs specify an amount of tokens and their owner. Blocks exhaust existing outputs and create new outputs with the same total amount of tokens, but with different distribution and owners. Each output can be spent only once: all future blocks that try to exhaust it will be rejected by the nodes. You can ask a node to go through all blocks for a list of created and exhausted outputs that mention you as an owner. If you strike out all exhausted outputs from the list of created outputs, you will get a list of *unspent outputs*. The sum of their tokens will make your token balance.
```

shimmer/learn/welcome.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Unlock Conditions, Addresses, and Seeds
+
+Each output has a list of unlock conditions. "Owning" an output means that it has an Address unlock condition that has your address listed. Addresses are not arbitrary: you can have multiple of them, but they all are derived from your seed through a special mathematical function.
```
samuel-rufi, line 30:
```suggestion
Each output has a list of unlock conditions. An output can only be unlocked by the owner of the address. Addresses are not arbitrary: you can have multiple of them, but they all are derived from your secret seed through a special mathematical function.
```

samuel-rufi (changes requested):

shimmer/learn/welcome.md
```diff
@@ -0,0 +1,50 @@
...
+
+You can also create a foundry to mint *native tokens*. Consider these tokens a custom currency. You can mix tokens from the same factory together.
+
+### Smart Contracts
+
+Shimmer supports smart contracts as a double-layered system. On the Layer 2, there are independent blockchains that actually run smart contracts. They regularly commit their state on the Layer 1, the Tangle. See [IOTA Smart Contracts](https://wiki.iota.org/smart-contracts/overview) for further detail.
+
+### Storage Deposits
+
+Shimmer is completely feeless. There are no fees for sending transactions, minting NFTs, or anchoring smart contract chains. But all this additional information has to be stored on the actual physical hardware. If not limited, that would cause node owners to add terabytes upon terabytes of additional storage space to their computers. For this reason, the available storage is bound to the amount of $SMR tokens in outputs. If there is some data in the output (an NFT or a list of native tokens), the $SMR tokens turn into its *storage deposit*. If during a transfer the new output has too little $SMR tokens, the data (with all the native tokens) will be lost.
```
samuel-rufi, line 48: > If during a transfer the new output has too little $SMR tokens, the data (with all the native tokens) will be lost.

Is this really true?

samuel-rufi (changes requested):

shimmer/learn/welcome.md
```diff
@@ -0,0 +1,50 @@
...
+
+Now, when the nodes agreed on the Tangle state, they can extract useful information from it: for example, who owns what tokens. Each block specifies *outputs*. In the most basic form, outputs specify an amount of tokens and their owner. Blocks exhaust existing outputs and create new outputs with the same total amount of tokens, but with different distribution and owners. Each output can be spent only once: all future blocks that try to exhaust it will be rejected by the nodes. You can ask a node to go through all blocks for a list of created and exhausted outputs that mention you as an owner. If you strike out all exhausted outputs from the list of created outputs, you will get a list of *unspent outputs*. The sum of their tokens will make your token balance.
```
samuel-rufi, line 26: Let's better point out the UTXO model, so users can look it up on google for better understanding. Instead of:
> Each block specifies *outputs*.

write something like

> According to the UTXO model, each block specifies *outputs*

samuel-rufi (changes requested):

shimmer/learn/welcome.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Basics
+
+### Nodes
+
+Shimmer is a network made out of *nodes*, computers that run the Shimmer node software. Anyone can join and leave the network, and nodes are run by individual parties. If you own an average computer, you can run your own Shimmer node, too.
+
+### Blocks and Outputs
+
+The structure that stores the data is called the [Tangle](./tangle.md). It consists of separate *blocks*, each of which stores some small data change. By including blocks into the Tangle, nodes agree on how the Tangle looks like. To get included, each block should reference from two to eight older blocks, and be referenced by a newer block. It means that there is always a front of the most recent blocks: you even can [see](https://explorer.shimmer.network/testnet/visualizer/) it. The more new pending blocks there are, the wider the front and the more points are there for newer blocks to attach. This scales really well.
```
samuel-rufi, line 24: > This scales really well. 

Lets remove that statement about scalability. There are multiple factors that impact scalability (e.g. consensus, ledger design, sharding, ...), lets give not a wrong impression.

A DAG, or blockDAG, cannot be considered a “solution” or “new protocol” in and of itself. Instead, a blockDAG is a framework for devising consensus protocols that may (or may not) be as secure as and more scalable than chain-based protocols.²
https://ancapalex.medium.com/an-introduction-to-the-blockdag-paradigm-50027f44facb

samuel-rufi, line 24: > each block should reference from two to eight older blocks
Is is important to be precise here? This value might change (not even sure its correct)

Can we just say "multiple"?

samuel-rufi (changes requested):

shimmer/learn/welcome.md
samuel-rufi, line 24: > By including blocks into the Tangle, nodes agree on how the Tangle looks like.

maybe this is better:

> By including blocks into the Tangle, nodes agree on the new ledger state.

samuel-rufi (changes requested):

PR #783, opened 2022-10-27
Fixes and formatting in wallet tutorial

adrian-grassl:
# Description of change

A few changes before this tutorial can be a proper prerequisite for further tutorials on e.g. NFT's (IPFS, Nested, Collection, ...)

## Type of change

Documentation Fix

## Change checklist

### DIFF
--- a/tutorials/single-page-tutorials/wallet-setup.md
+++ b/tutorials/single-page-tutorials/wallet-setup.md
@@ -173,34 +173,34 @@ This part of the code will:
 
 ```javascript
 async function run() {
-  	try {
-		// Define the account manager options with the imported network configuration and environment variables
-		const accountManagerOptions = {
-			storagePath: `./${accountName}-database`,
-			clientOptions: {
-				nodes: [nodeURL],
-				localPow: true,},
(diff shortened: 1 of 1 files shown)


jlvandenhout:
@Dr-Electron @lucas-tortora In the Wiki itself we converged on using two spaces instead of tabs for indentation. Prettier somehow doesn't seem to catch this in the codeblocks of this tutorial, while it should have. But do we care?

adrian-grassl:
@jlvandenhout It looks especially off in the github preview, but not that bad in the actual wiki render. However, tab size in my local IDE is set to 4 spaces. No problem with changing this to 2, if we aligned on that for the wiki.

jlvandenhout:
Let's keep it at 2 spaces then, to prevent a mix of tabs and spaces, and since other tutorials already seem to have converged to that standard.

jlvandenhout:
Still not sure why Prettier doesn't catch this though, it should be able to correctly format code blocks too, according to the language used. I'll see what I can do about that.

jlvandenhout (approved):

PR #786, opened 2022-11-04
Feat/nested nft tutorial

adrian-grassl:
# Description of change

Add tutorial to showcase the concept of nested NFTs on Shimmer.

## Type of change

Documentation Enhancement

## Change checklist


tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+    const StrongholdAccount = await manager.getAccount(accountName);
+    await StrongholdAccount.sync();
+
+    const ipfsNode = await startIpfsNode();
+
+    const items = ['avatar', 'mask', 'weapon'];
+
+    // This loop through the list of file names
+    for (const item of items) {
+      let cid = await uploadToIPFS(ipfsNode, `${item}.jpg`);
```
Dr-Electron, line 206:
```suggestion
      let cid = await uploadToIPFS(ipfsNode, `${item}.png`);
```

Again. You are using pngs :trollface:

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -217,12 +221,197 @@ Run the script `create-nfts.js` and check the console output to follow along the
...
 
-# Open ToDo's:
+Since you were minting three NFTs, your console output should look like this:
+
+```console
+Start local IPFS node for upload:
+<IPFS-node-swarm-information>
+
+
+Your file "avatar.jpg" was successfully uploaded to IPFS:
```
Dr-Electron, line 231: The actual message is different. You don't mention the file name

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -217,12 +221,197 @@ Run the script `create-nfts.js` and check the console output to follow along the
...
+
+Your NFT was successfully minted in this block:
+https://explorer.shimmer.network/testnet/block/<blockId>
+
+
+
+Your file "mask.jpg" was successfully uploaded to IPFS:
+https://ipfs.io/ipfs/<ipfs-content-identifier>
+
+Your NFT was successfully minted in this block:
```
Dr-Electron, line 242: For me it doesn't mint the second NFT.
It panics with 
```
Error:  {
  type: 'ClientError',
  error: "`The wallet account doesn't have enough balance. It only has 0, required is 60900`"
}
```

Not sure why tho. I also tired to add an additional sync() 🤔

Dr-Electron (changes requested):

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+## Storyline
+
+Let's imagine a game with avatars and items, which all are represented as unique NFTs on the Shimmer network. In this example it allows an avatar to not only hold items in-game, but its avatar NFT to also own its unique item NFTs directly on-DLT, in order to represent the dependencies between in-game components on the Shimmer network.
+
+![Blue Warrior](./images/warrior-blue.gif)
+
+## Prerequisites
+
+- [Node.js](https://nodejs.org/en/).
+- This tutorial will build on the tutorial [Testnet Wallet Setup](https://wiki.iota.org/tutorials/wallet-setup). If you haven't completed it, please do so, and afterward, come back here, or make sure you have a prefunded wallet. You will use the same folder and development environment as in the mentioned prerequired tutorial, allowing you to reuse the Stronghold account with funds on it.
```
jlvandenhout, line 14: I'd actually prefer this to be more decoupled. The basic requirement is to have a funded wallet right? So my suggestion would be to provide alternatives here on how to get one, and just point to the wallet tutorial as one of the alternatives.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "Apache-2.0"
+}
+```
+
+---
```
jlvandenhout, line 58: Referring to the previous comment, I'd set this up as a fresh project with its own dependencies.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+
+```javascript
+// Libraries
+const { utf8ToHex } = require('@iota/client');
+const { AccountManager } = require('@iota/wallet');
+const IPFS = require('ipfs-core');
+const fs = require('fs');
+
+// Environment variables
+require('dotenv').config({ path: './.env' });
```
jlvandenhout, line 92:
```suggestion
require('dotenv').config();
```

It looks for a `.env` file in the root of the project by default.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+const password = process.env.SH_PASSWORD;
+const accountName = process.env.ACCOUNT_NAME;
+
+// Network configuration
+const { networkConfig } = require('./networkConfig.js');
```
jlvandenhout, line 97: If we change the `networkConfig.js` module to just export the object directly, you don't need to unpack the export object.

```suggestion
const networkConfig = require('./networkConfig.js');
```

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+  console.log('\n');
+  console.log(consoleColor, `Start local IPFS node for upload:`);
+
+  let node;
+  if (!node) {
+    node = await IPFS.create({
+      repo: `ipfs_node`,
+    });
+  }
+  return node;
```
jlvandenhout, line 120: The if statement always evaluates to true, so unnecessary.

```suggestion
  return await IPFS.create({
    repo: `ipfs_node`,
  });
```

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+    const contentIdentifier = fileAdded.path;
+
+    console.log('\n');
+    console.log(consoleColor, `Your file was successfully uploaded to IPFS:`);
+    console.log(`https://ipfs.io/ipfs/${contentIdentifier}`, '\n');
+
+    return contentIdentifier;
+  } catch (error) {
+    console.error('IPFS upload error', error);
+  }
```
jlvandenhout, line 146: I know it is good practice to catch errors and handle them, but if we just print them to the console, I actually don't see any value in the extra code. Just let it crash, which will print the error to the console by default.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+}
+```
+
+#### 4. Mint NFT
+
+The function `mintNFT()` takes a NFT name, the previously created and unique IPFS identifier and an unlocked Stronghold account in order to mint an NFT on the configurated network. The metadata will be stored in the immutable metadata field, meaning it will remain unchanged for as long as the NFT itself exists on the network. Also note that the function does not pass a specific address to the Stronghold Account method `mintNfts()`. By default the method will mint all NFTs on the first address of the used Stronghold account, as long as it holds enough funds for Storage Deposit.
```
jlvandenhout, line 152: Referring to previous comments. In case we choose to not build on the wallet tutorial but make this tutorial more stand-alone, we actually do want to pass the address that the reader funded, one way or another.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+    uri: `https://ipfs.io/ipfs/${ipfsCid}`,
+  };
+
+  const metadataBytes = utf8ToHex(JSON.stringify(metadataObject));
+
+  const response = await StrongholdAccount.mintNfts([
+    {
+      immutableMetadata: metadataBytes,
+    },
+  ]);
```
jlvandenhout, line 178: We need to wait for the next milestone and sync to prevent the "no funds" issue.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+    const items = ['avatar', 'mask', 'weapon'];
+
+    // This loop through the list of file names
+    for (const item of items) {
+      let cid = await uploadToIPFS(ipfsNode, `${item}.jpg`);
+      await mintNFT(item, cid, StrongholdAccount);
+    }
+  } catch (error) {
+    console.log('Error: ', error);
+  }
```
jlvandenhout, line 211: Same as above, if we are just logging the error, I'd just let the default error handling do its thing and remove the try/catch block.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+  process.exit(0);
```
jlvandenhout, line 212: What is the reason for this?

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+:::
+
+#### 1. Imports and parameters
+
+```javascript
+// Libraries
+const { AccountManager } = require('@iota/wallet');
+
+// Environment variables
+require('dotenv').config({ path: './.env' });
```
jlvandenhout, line 333:
```suggestion
require('dotenv').config();
```

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+const password = process.env.SH_PASSWORD;
+const accountName = process.env.ACCOUNT_NAME;
+
+// Network configuration
+const { networkConfig } = require('./networkConfig.js');
```
jlvandenhout, line 338:
```suggestion
const networkConfig = require('./networkConfig.js');
```

Same as above if we refactor the `networkConfig` module.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+      nftId: '<insert-id-of-nft-you-want-to-send>',
+    }]);
+
+    console.log(consoleColor, `Your NFT was successfully sent to the target address in this block:`);
+    console.log(`${explorerURL}/block/${response.blockId}`, '\n');
+
+  } catch (error) {
+      console.log('Error: ', error);
+  }
+  process.exit(0);
```
jlvandenhout, line 371: Same as above, I guess we can do without all of this.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
```diff
@@ -0,0 +1,423 @@
...
+```text
+avatar NFT Address:
+rms...
+
+mask NFT ID:
+0x...
+
+weapon NFT ID:
+0x...
+```
```
jlvandenhout, line 314: This whole section kind of feels clumsy and hard to follow. Can't we just get this info directly from the response or from the explorer?

jlvandenhout (changes requested):

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
adrian-grassl: Depends on how you define wallet. In this case the subsequent tutorials require a Stronghold account within the same folder in order to work. So just a mnemonic with funds on it is not sufficient.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
adrian-grassl: Maybe we can discuss this in our call today. In case we want to establish learning paths, IMO these kind of dependencies shouldn't be problematic. In case we want every tutorial to be independent you're right of course.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
adrian-grassl: I added a function to await transaction confirmation before the next NFT is minted. Probably not the most elegant way, but gets the job done.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
adrian-grassl: Agreed. But since this has dependencies to the wallet setup tutorial, I'll keep it this way to avoid any breaks until we've aligned on if we want any tutorial to be independent, or if they are connected learning paths.

tutorials/single-page-tutorials/nested-nfts-with-wallet-lib.md
Dr-Electron: Yes that should be possible by using the minting result and storing the IDs 🤔

Dr-Electron:
Hi Adrian does the tutorial still work for you? I tried it but by using the cli-wallet to get funds, but it seems to me like the code is outdated 😅 

Btw, the rust of trying until confirmed is now also part of node.js https://github.com/iotaledger/wallet.rs/pull/1632

PR #795, opened 2022-11-19
Refactor `src`

jlvandenhout:
# Description of change

This refactors the `src` directory to have a more logical separation between the different environments.

## Type of change

- Enhancement (a non-breaking change which adds functionality)

## Change checklist

Tick the boxes that are relevant to your changes, and delete any items that are not.

- [x] I have followed the [contribution guidelines](https://github.com/iota-wiki/iota-wiki/blob/main/.github/CONTRIBUTING.md) for this project
- [x] I have performed a self-review of my own changes
- [x] I have made sure that added/changed links still work
- [ ] I have commented my code, particularly in hard-to-understand areas

PR #796, opened 2022-11-19
Refactor configs

jlvandenhout:
# Description of change

STACKED ON TOP OF #795

This refactors the configuration throughout the repo. Configuration files are now globbed and deep merged properly. This allows you to include, exclude, and override certain settings, depending on in what order you merge the configurations (configurations included later in the merge override previous configurations where they overlap). This allows for example to create development and production configurations more easily.

Next to the existing CONTENT environment variable, we now also have an ENVIRONMENT environment variable that selects either the IOTA, Shimmer or Next environment. Package scripts and workflows are updated accordingly.

## Type of change

## Change checklist


begonaalvarezd (commented):
@jlvandenhout we've been using this branch (`refactor/configs`) since yesterday because the build time is way faster than before, but we just realized that not all endpoints are available, for some reason docusaurus is throwing 404 in urls that should exist, eg `/shimmer/iota.rs/welcome`

jlvandenhout:
Just for future reference: In the current state of the repo we can run dev servers for each environment individually, but we are unable to start dev servers for the 3 environments at the same time, because Docusaurus uses the `.docusaurus` directory by default, so the dev server builds overwrite each other. To run them in parallel we have to do a full build for each of them and serve the result, which takes a long time, and even with external documentation included. This is why we have the `CONTENT` environment variable to select what external documentation we want to include (it accepts comma separated glob patterns relative to the `<environment>/external` directories). For convenience we provide `yarn start:all`, which does a full build including all external documentation.

Dr-Electron (approved):

PR #800, opened 2022-11-28
feat: add google analytics

marc2332:
# Description of change

Added Google Analytics. 

Live deploy: https://iota-wiki-mjwdlmc2m-mkenzo8.vercel.app/

Fixes https://github.com/boxfish-studio/iota-wiki/issues/3

## Type of change

- Enhancement

## Change checklist


jlvandenhout:
We could also put the configuration in the root `docusaurus.config.js` [here](https://github.com/iota-wiki/iota-wiki/blob/41d949b0cc0a03f67a935e6206e42e74ff819fbb/docusaurus.config.js#L29), saves us an extra config file and having to import it in the 3 environment configs. wdyt?

marc2332:
> We could also put the configuration in the root `docusaurus.config.js` [here](https://github.com/iota-wiki/iota-wiki/blob/41d949b0cc0a03f67a935e6206e42e74ff819fbb/docusaurus.config.js#L29), saves us an extra config file and having to import it in the 3 environment configs. wdyt?

That was actually my initial goal but I couldn't seem to get it working with this `merge function that is being used. It would definitely be a better way.

jlvandenhout:
Ah, I see you add it to the preset config. I wanted to move away from using the preset anyway (we use less and less of it), so I'm fine adding this directly as a plugin. In this case in `docusaurus.config.js` we can add it to the production plugsin array:

```
const production = {
  ...
  plugins: [
    ...
    [
      '@docusaurus/plugin-google-gtag',
      {
        trackingID: '<tracking_id>',
        anonymizeIP: true,
      },
    ],
  ],
};
```

begonaalvarezd (changes requested):
After the latest changes it doesnt seem working anymore

begonaalvarezd (approved):
Now works, it needs to be triggered with a `build:production` and now it appears 🙌🏼

PR #858, opened 2023-01-19
DRAFT - Redo develop section

lucas-tortora:
# Description of change

Redid the develop section to better suit an education plan and reduce pain points to reach dev docs.

## Open Questions

1. Should we add links or imports for the `start building` section?
3. Should we feature a single client lib or add examples in all libs?
4. Should we feature a particular language or set of languages?
5. If we choose to feature multiple libs and/or languages, we need to discuss the structure:
    1. Use an action-oriented sidebar and add headers or tabs for libs and tabs for languages
    
image

    2. Use the library as a sidebar category > action as file and tabs for languages
    
image

    3. Use the language as sidebar category > action as file and tabs for libs?
    
image

## Type of change

- Documentation Enhancement

## Change checklist


Dr-Electron:
Ok, this is much to handle at once. Let me try:

> 1. Should we add links or imports for the start building section?
What do you mean with links or imports?

> 2. Should we feature a single client lib or add examples in all libs?
With feature you mean what exactly? Why even add examples?
If we go for the separation you are suggesting here, wouldn't the `Start Building` section focus on the "learn" part of developers. So stuff like differences, general best practices on handling secrets and so on.
Examples would still be in the external docs, no?

> 3. Should we feature a particular language or set of languages?
Same here.

> 4. If we choose to feature multiple libs and/or languages, we need to discuss the structure:
And same here. We wanted to keep docs in the repo so it is easier for the corresponding people (internal devs, external devs) to contribute and keep up to date. If we keep stuff like installation guides in the wiki this will be a nightmare for us to maintain 🤔

aleksei-korolev:
1. We should explicitly refer the user to the external articles, otherwise the readers can't learn which part of the wiki has which kind of content, and that will hinder the navigation in the future.
2. It depends on their goal, what kind of problems people have when they start develop with IOTA and what is usually enough to solve them? The combination of iota.rs+wallet.rs? @adrian-grassl 
3. Even though I'd prefer to deal with just one language, I assume that JS is much more popular than Rust, and I would keep Python just because everyone understands Python even if they never seen Python code before. We can maintain the code snippets for other languages, but I would ditch them the moment they give us trouble.
4. The action comes first, the language comes second. We did it in the dev docs, it works very well, the other way around worked poorly. I see no reason why we should do it the other way around here.

@adrian-grassl do you maybe have done a research before, what kind of topics/concepts/skills a person must have before he can orient himself in the rest of the ecosystem/documentation?

lucas-tortora:
Thanks for the feedback @Dr-Electron and @aleksei-korolev !! I'm incredibly glad the consensus is using links cause the embeds road looked promising at first but got really bumpy really fast.

adrian-grassl:
@aleksei-korolev I'm afraid we don't have that one developer skill profile or library which would cover 90% of cases out there. I am however quite convinced, that if we do a good job in the layers above (What is our ecosystem? Which networks are there and which one is best for me? And so on...) devs will figure out which lib is best for them quite fast.

PR #881, opened 2023-02-20
Fix client naming

Dr-Electron:
# Description of change

I'm sorry for having added this to the wiki :(

### DIFF
--- a/next/develop/welcome.md
+++ b/next/develop/welcome.md
@@ -11,9 +11,9 @@ Welcome to the Developer Section. Here you will find all information to start de
   - [WASP CLI](/smart-contracts/guide/chains_and_nodes/wasp-cli)
   - [Schema Tool](/smart-contracts/guide/wasm_vm/schema)
 - Libraries
-  - [Client.rs](/iota.rs/welcome)
+  - [iota.rs](/iota.rs/welcome)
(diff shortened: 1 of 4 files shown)


next/develop/welcome.md
```diff
@@ -11,9 +11,9 @@ Welcome to the Developer Section. Here you will find all information to start de
   - [WASP CLI](/smart-contracts/guide/chains_and_nodes/wasp-cli)
   - [Schema Tool](/smart-contracts/guide/wasm_vm/schema)
 - Libraries
-  - [Client.rs](/iota.rs/welcome)
+  - [iota.rs](/iota.rs/welcome)
   - [Wallet.rs](/wallet.rs/welcome)
-  - [Client.js](/iotajs/welcome)
+  - [iota.js](/iotajs/welcome)
```
lucas-tortora, line 16: Probably out of scope for this PR, but we may want to define if we use the `.` in the URLs. I doubt many people input them by hand, but it looks weird to use it for `.rs` and not `.js`

lucas-tortora (approved):
LGTM

next/develop/welcome.md
Dr-Electron: Sure we can discuss it. What do you prefer? Probably easier would be to add the `.` to iota.js as less people use those docs 😅

PR #922, opened 2023-04-07
Add rehype-lodash-template to use template strings

### LINKED ISSUE #879: Add templates for version replacements

To minimise the problems with merging different version, we should provide variables that replace certain templates so that merging docs won't be a problem (see examples at the end of this issue)

**Describe the solution you'd like**

We maybe could use [rehype-lodash-template](https://github.com/viktor-yakubiv/rehype-lodash-template) or a similar tool to have a list of templates/variables that in the docs will be replaced with the corresponding strings.

For urls this may also be of help: [rehype-url-inspector](https://github.com/JS-DevTools/rehype-url-inspector) (It could even help with [broken links](https://github.com/iota-wiki/iota-wiki/issues/880))

**Describe alternatives you've considered**

Doing it manually :trollface: 

**Additional context**

In the docs you could for example use `${ENVIRONMENT_URL}/hornet/welcome` which in a Shimmer environment would point to `https://wiki.iota.org/shimmer/hornet/iota.org` (or even better use the config parameters) and in an IOTA environment to `https://wiki.iota.org/hornet/iota.org`

Or another markdown example:
```
In ${NETWORK} you always need a storage deposit
```

where `${NETWORK}` will be replaced with either `IOTA` or `Shimmer`

Dr-Electron:
# Description of change

This should make it possible to use template strings like `${network}` for easier management of different versions of our docs

Fixes #879 

## Type of change

## Change checklist

### DIFF
--- a/iota/docusaurus.config.js
+++ b/iota/docusaurus.config.js
@@ -1,6 +1,6 @@
 const path = require('path');
 
-async function createConfig() {
+module.exports = async () => {
   const rehype_lodash_template = (await import('rehype-lodash-template')).default;
 
   return {
@@ -290,5 +290,3 @@ async function createConfig() {
     ],
   };
 }
-
-module.exports = createConfig;
--- a/iota/docusaurus.config.js
+++ b/iota/docusaurus.config.js
@@ -1,7 +1,8 @@
 const path = require('path');
 
 module.exports = async () => {
-  const rehype_lodash_template = (await import('rehype-lodash-template')).default;
+  const rehype_lodash_template = (await import('rehype-lodash-template'))
+    .default;
 
   return {
     title: 'IOTA Wiki',
@@ -229,7 +230,7 @@ module.exports = async () => {
                 values: {
                   network: 'IOTA',
                 },
-              }
+              },
             ],
           ],
           showLastUpdateTime: true,
@@ -289,4 +290,4 @@ module.exports = async () => {
       ],
     ],
   };
-}
+};
--- a/iota/docusaurus.config.js
+++ b/iota/docusaurus.config.js
@@ -1,280 +1,294 @@
 const path = require('path');
 
-module.exports = {
-  title: 'IOTA Wiki',
-  tagline: 'The complete reference for IOTA',
-  baseUrl: '/',
-  themeConfig: {
-    announcementBar: {
-      id: 'govern',
-      content:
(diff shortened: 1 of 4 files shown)


jlvandenhout (approved):

Dr-Electron:
I have some open questions on this one:

- [ ] We would add this to every plugin separately, right? So every repo could also configure their own templates.
- [ ] Sadly we can't really use this for links. For example replacing wiki environments would be nice. Something like:
```js
remarkPlugins: [
    [
        rehype_lodash_template,
        {
            values: {
                wiki-base-url: 'https://wiki.iota.org/shimmer/',
            },
        }
    ],
],
```
Which would allow to write text like:
```
You can find more info in this Shimmer docs [here](${wiki-base-url}some/path/in/the/shimmer/wiki)
```
But if we do that, our current broken link checker won't work as it works on the markdown AST. 
Not sure if we can somehow run the check after the remark plugins were run. Or if we just need to find a rehype plugin that checks for broken links 🤔 

Wdyt @lucas-tortora and @jlvandenhout .
But after all I think the tool is soo nice that it makes sense to merge anyway. Just something we probably need to handle in the future

lucas-tortora (approved):

lucas-tortora:
> * [ ]  We would add this to every plugin separately, right? So every repo could also configure their own templates.

Can we do both? Each plugin would load the wiki templates, which could be complemented and/or overridden.

Dr-Electron:
> Can we do both? Each plugin would load the wiki templates, which could be complemented and/or overridden.

Would be nice. But not sure what the best solution would be 🤔 
We could set env variables with the cli or something like that 🤔

jlvandenhout:
> We would add this to every plugin separately, right?

Like it is implemented now, we need to really. We can't have a "global remark plugin", as remark plugins act on individual docs instances.

> So every repo could also configure their own templates.

This is a direct consequence of the above, not necessarily a choice.

What we could do is provide a default mapping from the core package or theme package, that we can import wherever we have a use for these templates.

An alternative is to wrap the remark plugin ourselves and provide a default mapping in the wrapping plugin. That way we can implement logic to make it work nice with the rest of the Wiki build system, like having `network` mapped to the configured `ENVIRONMENT` environment variable, or check that default mappings are not shadowed by project defined mappings, etc.

> Sadly we can't really use this for links

IMO this is not necessarily an issue. We should really take some time and figure out and document how to link properly in the Wiki. Because normally you wouldn't have to use tricks like this to properly link for example between pages under the `shimmer` subpath with relative URLs.

Issue #904, opened 2023-03-26
"Add Custom Network" button for ShimmerEVM


huhn511:
**Describe the solution you'd like**
Add a "Add Custom Network" button or link to add the ShimmerEVM with a single click to MetaMask. 

This can be added for example as link in the footer or as a button to all tutorials, which uses the ShimmerEVM.

## Example Code:
Here is a code snippet for a button that, when clicked, adds a custom network to MetaMask:
**View**:
```js
  <button onclick="addNetwork()">Add Custom Network to MetaMask</button>
```
**Controller:**
```js
async function addNetwork() {
      if (typeof window.ethereum !== 'undefined') {
        try {
          await window.ethereum.request({
            method: 'wallet_addEthereumChain',
            params: [
              {
                chainId: '0x89', // Replace with the chain ID of the network you want to add
                chainName: 'Custom Network', // Replace with the name of the network
                nativeCurrency: {
                  name: 'Custom Token', // Replace with the name of the native currency
                  symbol: 'CTK', // Replace with the symbol of the native currency
                  decimals: 18, // Replace with the number of decimals of the native currency
                },
                rpcUrls: ['https://custom-rpc-url'], // Replace with the RPC URL(s) of the network
                blockExplorerUrls: ['https://custom-blockexplorer-url'], // Replace with the block explorer URL(s) of the network (optional)
              },
            ],
          });
        } catch (error) {
          console.error(error);
          alert('Error adding network: ' + error.message);
        }
      } else {
        alert('MetaMask is not installed. Please install MetaMask and try again.');
      }
    }
```

Replace the placeholder values in the params object with the details of the network you want to add to MetaMask. When a user clicks the "Add Custom Network to MetaMask" button, this code snippet will attempt to add the specified network to the user's MetaMask extension. If MetaMask is not installed, it will display an alert prompting the user to install MetaMask.

Dr-Electron:
Duplicate of https://github.com/iota-wiki/iota-wiki/issues/882

Dr-Electron:
For this to work we need to add it to https://github.com/iota-wiki/shared-components so everyone (wiki and external repos) can use it. 
But maybe we should also discuss to add shared components to the wiki. I like the monorepo wiki we have now 🤔 . And maybe at some point we want to "deliver" those components together with the wiki (because we only need the components temporarily in the external repos anyway) and maybe that would be easier if everything is part of the same wiki.
[...]

PR #925, opened 2023-04-15
Feat/add CodeSnippetBlock theme component

jlvandenhout:
# Description of change

This adds the `@iota-wiki/theme` package, exposing the CodeSnippetBlock as first implemented component. This way, the theme components can be used in the Wiki build, but also used in local builds of documentation using the IOTA Wiki CLI. Examples of how to use the exposed components are provided in the `theme/README.md` file and in the CLI tests. To see the CodeSnippetBlock in action:

- `yarn && yarn ./cli build` to build the CLI.
- `yarn iota-wiki start cli/tests/tutorial/` to start the Tutorial test.
- Navigate to `/docs/tutorial-basics/markdown-features/#codesnippetblock` to see the component rendered.

iotaledger/guild-devx#3
#904

## Type of change

## Change checklist


jlvandenhout:
@Dr-Electron @lucas-tortora could you have a look at the open suggestions/questions?

Dr-Electron:
Nice work. I would definitly change the logic to ignore our „special comments“. This could come in handy to just mark relevant stuff with comments and it probably will be less error prone.

The empty options would also be nice. But not a top priority. If we can just implement it quickly, go for it

jlvandenhout:
I changed the logic to include only the lines between the matched beginning and end, excluding the matching lines themselves and any surrounding whitespace. Both the start string and end string can be omitted, in which case the start and end of the file are taken respectively.

I chose to allow the situation where a `startString` and/or `endString` attribute is present, but cannot be found in the code. In this case the start and end of the file are taken respectively. This is to prevent a small typo or change in example code from breaking the entire Wiki build.

lucas-tortora (approved):
Awesome work! Thanks @jlvandenhout !! My only thought is that perhaps we want to break the wiki build if the string isn't found, so we know that it needs our attention, but we can certainly update that behaviour later if needed

Dr-Electron (approved):

Dr-Electron:
> Awesome work! Thanks @jlvandenhout !! My only thought is that perhaps we want to break the wiki build if the string isn't found, so we know that it needs our attention, but we can certainly update that behaviour later if needed

Yeah good idea. You should create an issue for it :trollface:

Dr-Electron:
And @jlvandenhout when use your own tools to format? :trollface:

Dr-Electron (approved):

PR #934, opened 2023-04-21
feat: search for similar urls in network sitemap

VmMad:
# Description of change

Update network switcher component to find in the network sitemap for a similar url or a fallback. In case there aren't any, then redirect to a `/404`.

## Links to any relevant issues

Fixes #781

## Type of change

- Bug fix

## Change checklist

- [x] I have followed the [contribution guidelines](https://github.com/iota-wiki/iota-wiki/blob/main/.github/CONTRIBUTING.md) for this project
- [x] I have performed a self-review of my own changes
- [x] I have made sure that added/changed links still work
- [x] I have commented my code, particularly in hard-to-understand areas

PR #935, opened 2023-04-21
feat: add script to watch user clicks on search page

### LINKED ISSUE #794: Search results are broken
**Describe the bug**
The search results are broken https://wiki.iota.org/search?q=native%20token
All links there return `Page Not Found`

**To Reproduce**

1. Go to https://wiki.iota.org/search?q=native%20token
2. Click on a link
3. See error

**Expected behavior**
Results should go to a working page

**Desktop (please complete the following information):**

- OS: [pop!_os]
- Browser [ chrome]

VmMad:
# Description of change

Add a script to watch if a user clicks on a link in the search page to avoid navigating with the SPA method.

Fixes #794 

## Type of change

## Change checklist

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -26,6 +26,10 @@ module.exports = async () => {
 
   const internal = await glob(['tutorials', 'common', ENVIRONMENT]);
 
+  const scripts = {
+    scripts: ['/js/search-environment-observer.js'],
+  };
+
   const search = {
     themeConfig: {
       algolia: {
@@ -65,5 +69,6 @@ module.exports = async () => {
     ...internal,
     search,
     isProduction ? production : {},
+    scripts,
   );
 };
--- a/static/js/search-environment-observer.js
+++ b/static/js/search-environment-observer.js
@@ -1,33 +1,33 @@
 function handleOnClick(e) {
-    const isUserInSearchPage = window.location.pathname.includes('/search/');
-    const targetHref = e.target.href;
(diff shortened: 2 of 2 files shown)


Dr-Electron (approved):
LGTM. Although I have a similar problem here like with https://github.com/iota-wiki/iota-wiki/pull/934.
I'm afraid that with the "special" detection of next and shimmer if we ever update base urls or something we will forget to update this files 😅 .
Not sure if there would be a better way to share configs somehow tho 😅

jlvandenhout:
On the other hand, I'm fine with merging this, as it allows the "initial search experience" to be inside the environment, while if the reader doesn't find what it is looking for, it can click on the "show more results" and it shows results outside the environment too, which is kind of nice I guess.

jlvandenhout:
> LGTM. Although I have a similar problem here like with #934. I'm afraid that with the "special" detection of next and shimmer if we ever update base urls or something we will forget to update this files 😅 . Not sure if there would be a better way to share configs somehow tho 😅

And could you open an issue for this? I think this is something that is fixable.

Dr-Electron:
I actually think the PR fixes the linked issue. But there should be another issue on how we handle different environments.
Didn't we talk about sorting them. So if you search in IOTA it would also show you Shimmer results, just not on the first page.
Maybe even with different theming. Or a dropdown in the search. I remember a lot of ideas in the room.

Dr-Electron:
> And could you open an issue for this? I think this is something that is fixable.

Will do

PR #963, opened 2023-05-24
restart cli preview server when root config changes

### LINKED ISSUE #862: Make the cli use the most up to date config(s)

Right not the latest cli release doesn't for example use our current trailing slash config.

**Describe the solution you'd like**
It would be nice if the cli could just pull the config from GitHub or something similar. If we do that we may just need to make sure that we also check if those paths change so that we know we need to release a new cli.

**Describe alternatives you've considered**
We could release a new cli every time we change something but I this doesn't really solve the problem in a nice way

VmMad:
# Description of change

Use nodemon to watch for changes in the working directory docusaurus.config.js.

## Installation steps

1. Install dependencies: `yarn`.
2. Clear cli files, install dependecies and build: `cd cli && yarn clean && yarn && yarn build && cd ..`
3. Start the cli: `yarn iota-wiki` or if that didn't work `yarn ./cli iota-wiki start 'path/to/test/project'` 
e.g. `yarn ./cli iota-wiki start 'test/module'`

Fixes #862 

## Type of change

- Enhancement

## Change checklist


VmMad:
Comment for reviewers:

This fix only works for changes made in the root `docusaurus.config.js`, but will ignore any changes made into imported files, for example changing something from `common/docusaurus.config.js`. 
If an implementation of this logic is needed, iI think it would be better to address that in another issue/PR, because it might not be as simple as it also happens when you run `yarn start:iota` and change something from `iota/docusaurus.config.js`. Importing from other files is not handled right now and I don't think is a cli only problem.

PR #1026, opened 2023-07-09
Use deploy-to-vercel-action instead of vercel cli

### LINKED ISSUE #1017: Use deploy-to-vercel-action action instead of vercel cli directly

Instead of doing more custom and manual workflows I think using https://github.com/BetaHuhn/deploy-to-vercel-action would make our workflow nicer

Dr-Electron:
# Description of change

I switched to this action to make the setup more straight forward

Fixes #1017 

## Type of change

## Change checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -37,15 +37,15 @@ jobs:
           tar -xf iota.tar
           tar -xf shimmer.tar
           tar -xf next.tar
-      # Convert to Vercel project and upload
-      - name: Install Vercel CLI
-        run: npm install --global vercel@latest
-      - name: Pull Vercel Environment Information
-        run: vercel pull --yes --environment=production --token=${{ secrets.VERCEL_TOKEN }}
-      - name: Build Project Artifacts
-        run: vercel build --prod --token=${{ secrets.VERCEL_TOKEN }}
-      - name: Deploy Project Artifacts to Vercel
-        run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
+      # Deploy to Vercel
+      - name: Deploy to Vercel Action
+        uses: BetaHuhn/deploy-to-vercel-action@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+          PR_LABELS: false
       # Trigger Algolia crawler
       - name: Crawl
         run: |
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -51,29 +51,12 @@ jobs:
           tar -xf shimmer.tar
           tar -xf next.tar
       # Convert to Vercel project and upload
-      - name: Install Vercel CLI
-        run: npm install --global vercel@latest
-      - name: Pull Vercel Environment Information
-        run: vercel pull --yes --environment=preview --token=${{ secrets.VERCEL_TOKEN }}
-      - name: Build Project Artifacts
-        run: vercel build --token=${{ secrets.VERCEL_TOKEN }}
-      - name: Deploy Project Artifacts to Vercel
-        run: vercel deploy --prebuilt --token=${{ secrets.VERCEL_TOKEN }} > vercel_output.txt
-      # Set the deploy URL as an output
-      - name: Set deploy url
-        id: deploy_url
-        run: echo "DEPLOY_URL=$(cat vercel_output.txt | awk 'END{print}')" >> $GITHUB_OUTPUT
-      # Create a comment on the pull request with the deployed URL
-      - name: Comment on pull request
-        uses: actions/github-script@v6.4.0
+      - name: Deploy to Vercel Action
+        uses: BetaHuhn/deploy-to-vercel-action@v1
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          script: |
-            const DEPLOY_URL = '${{ steps.deploy_url.outputs.DEPLOY_URL }}'
-            const COMMIT_SHA = '${{ github.event.pull_request.head.sha }}'
-            github.rest.issues.createComment({
-              issue_number: context.issue.number,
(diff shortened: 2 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -37,15 +37,15 @@ jobs:
...
-        run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
+      # Deploy to Vercel
+      - name: Deploy to Vercel Action
+        uses: BetaHuhn/deploy-to-vercel-action@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+          PR_LABELS: false
```
jlvandenhout, line 48: I guess this also needs a prebuilt flag?

```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
          PR_LABELS: false
          PREBUILT: true
```

jlvandenhout (changes requested):
I have no clue anymore why I did it the way I did, but if this now works, hey, it works!

.github/workflows/deploy.yml
Dr-Electron: Yeah I didn't completely understand why... 😅 
Do you know?

Dr-Electron:
> I have no clue anymore why I did it the way I did, but if this now works, hey, it works!

I think this was done by Begoña. So you are fine :trollface:

.github/workflows/deploy.yml
jlvandenhout: I think because normally you would provide the build steps to the Vercel config and let Vercel take care of everything, but we have a more complex build with 3 envs and different levels of caching, so we built manually and then just tell Vercel to deploy.

.github/workflows/deploy.yml
```diff
@@ -37,15 +37,16 @@ jobs:
...
+          PREBUILT: true
```
VmMad, line 49: This is not going to work because neither `FORCE` or `PREBUILT` are in the [latest released version](https://github.com/BetaHuhn/deploy-to-vercel-action/blob/v1.9.12/action.yml#L5C1-L77) of this action.
And judging by the workflow it looks like it might be an important flag. Just letting you know.

.github/workflows/deploy.yml
```diff
@@ -37,15 +37,16 @@ jobs:
...
-        run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
+      # Deploy to Vercel
+      - name: Deploy to Vercel Action
+        uses: BetaHuhn/deploy-to-vercel-action@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+          PR_LABELS: false
```
VmMad, line 48: I think it would be good to add `PRODUCTION: true` here just in case

.github/workflows/preview.yml
```diff
@@ -51,29 +51,12 @@ jobs:
...
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              body: `This pull request has been deployed to Vercel.\n\n\nLatest commit: ${COMMIT_SHA}\n\n:white_check_mark: Preview: ${DEPLOY_URL}`
-            })
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
+          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+          PR_LABELS: false
+          PREBUILT: true
```
VmMad, line 62: It would be good to add `PRODUCTION: false` here just in case.
Also, If you want the deploys to show in the `Preview` environment, then you should add `GITHUB_DEPLOYMENT_ENV: Preview`. We already had problems because we didn't add this line :sweat_smile:

VmMad (commented):
hey @Dr-Electron! I was reviewing the workflow and I don't know if using this action right now is the best idea. I left some comments:

.github/workflows/deploy.yml
Dr-Electron: Good catch. Thanks for that. We can just wait for the next release or add the latest changes from the repo.
@jlvandenhout wdyt?

.github/workflows/deploy.yml
Dr-Electron: Should be on by default, but we can be explicit 👍 
```suggestion
          PR_LABELS: false
          PRODUCTION: true
```

.github/workflows/preview.yml
Dr-Electron: Not familiar enough with Github and Action to tell what exactly that does 😅 
But I can add it 👍 
```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
          VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
          VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
          PR_LABELS: false
          PREBUILT: true
          PRODUCTION: false
          GITHUB_DEPLOYMENT_ENV: Preview
```

.github/workflows/deploy.yml
jlvandenhout: Yeah, after the restructure we will have a much less complex build, so we might go back to a more vanilla Vercel workflow anyway.

Dr-Electron: Ok. Will convert this to a DRAFT in the meantime. @VmMad thanks for your help ❤️

PR #1038, opened 2023-07-14
Update deploy and preview workflow

Dr-Electron:
# Description of change

Updated the deploy and preview workflow to make it work with the restructure

## Type of change

## Change checklist

- [ ] I have made sure that added/changed links still work


.github/actions/build/action.yaml
```diff
@@ -0,0 +1,48 @@
...
+    # Artifacts cached after install to prevent removal by Yarn
+    - name: Cache build artifacts
+      uses: actions/cache@v3
+      with:
+        path: node_modules/.cache
+        key: wiki-build-artifacts-${{ hashFiles('package.json', 'yarn.lock', 'src/**/*.*') }}
+    # Create reference docs
+    #- name: Build reference docs
+    #  run: |
+    #    ./scripts/gen_${{ inputs.environment }}_references.sh
```
jlvandenhout, line 44: Just to check, these were commented out for a reason right?

jlvandenhout: Because I don't see the refs being built anywhere else.

.github/workflows/dependabot-build.yaml
```diff
@@ -6,10 +6,8 @@ on:
...
     if: github.actor == 'dependabot[bot]'
-    strategy:
-      matrix:
-        environment: [iota, shimmer, next]
-    uses: ./.github/workflows/build.reusable.yaml
-    with:
-      environment: ${{ matrix.environment }}
+    steps:
+      - name: build
+        uses: ./.github/actions/build/
```
jlvandenhout, line 13: Do we need a checkout step before this? How else will it find the composite action?

.github/workflows/deploy.yml
```diff
@@ -8,35 +8,14 @@ on:
...
-        with:
-          name: build
-      # Paths of the build artifacts are preserved (`build`, `build/shimmer`, and `build/next`)
-      - name: Extract build
-        run: |
-          tar -xf iota.tar
-          tar -xf shimmer.tar
-          tar -xf next.tar
+      - name: build
+        uses: ./.github/actions/build/
```
jlvandenhout, line 18: I guess same as above, do we need a checkout step so the action can find the referenced composite action?

jlvandenhout (changes requested):

.github/actions/build/action.yaml
Dr-Electron: Good point. I commented them out as a quick fix, but I need to read them.

.github/workflows/dependabot-build.yaml
Dr-Electron: True that. Thanks

.github/workflows/deploy.yml
Dr-Electron: Correct

Dr-Electron: Fixed in b2f6dc5

.github/workflows/dependabot-build.yaml
Dr-Electron: Fixed in b2f6dc5

jlvandenhout (approved):

PR #1093, opened 2023-08-12
Swizzle version banner

### LINKED ISSUE #1087: Swizzle Versioning banners

We should swizzle(?) the versioning banner to mark our docs as for example deprecated.
But maybe it would be easier to create a new component and wrap the docs content. That way in the plugin we could specify a custom text even.

jlvandenhout:
# Description of change

This changes the version banner to show a deprecation caution when 'unmaintained' is used as banner value for that version. It shows a link to the latest version, if available and not the same as the deprecated version.

fixes #1087 

## Type of change

## Change checklist

### DIFF
--- a/contentPlugins.js
+++ b/contentPlugins.js
@@ -177,7 +177,7 @@ module.exports = async () => {
         current: {
           label: '1.1.0',
           badge: true,
-          banner: 'unmaintained',
+          banner: 'deprecated',
         },
       },
     },
--- a/contentPlugins.js
+++ b/contentPlugins.js
@@ -173,6 +173,13 @@ module.exports = async () => {
         __dirname,
         'docs/build/stronghold.rs/1.1.0/sidebars.js',
       ),
+      versions: {
+        current: {
+          label: '1.1.0',
+          badge: true,
+          banner: 'unmaintained',
+        },
+      },
     },
     {
       id: 'iota-streams',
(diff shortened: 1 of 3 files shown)


jlvandenhout:
@Dr-Electron would this suffice for now?

Dr-Electron:
Maybe that's already too complicated, but what if we just allow custom banners?
That way every project could add custom banners with content like: `unreleased`, `unmaintained` and more.

eike-hass:
Awesome stuff 💪 I think ideally the the banners would be flexible what content/text they display so they can be used many (including not anticipated) use-cases,

jlvandenhout:
> Maybe that's already too complicated, but what if we just allow custom banners? That way every project could add custom banners with content like: `unreleased`, `unmaintained` and more.
> I think ideally the the banners would be flexible what content/text they display so they can be used many (including not anticipated) use-cases

Yes, that would be ideal of course, but "just allow custom banners" is not really how Docusaurus thought of it, they hardcoded their banner options and validate the config as such too. So "not too complicated" and "flexible banners" would be on the opposite sides of the spectrum. Let me think this through for a bit.

Dr-Electron:
Maybe we should ask them to change it ;)

jlvandenhout:
Closing in favor of #1103

PR #1162, opened 2023-09-07
Add version redirects to current version

### LINKED ISSUE #1086: Allow "deep" links to main version of versioned project

Currently we link to the stable release of for example identity with /identity.rs/...
But would be nice if /identity.rs/v0.6 would link to the same version.

A possible solution could be to use the[ client redirect plugin](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects/). Not sure if this would be dependent on #1083.

Dr-Electron:
# Description of change

This PR adds version redirects so that the current version is not only accessible from `/<project>/**` but also from `/<project>/<current-version>/**`

Fixes #1086.

## Type of change

## Change checklist

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -375,6 +375,16 @@ module.exports = async () => {
                 },
               ];
 
+              // Version redirects are only used to asign path with the actual version it to the "current" version
+              const versionRedirects = [
+                {
+                  from: '/identity.rs/0.6',
+                  to: '/identity.rs',
+                },
+              ];
+
+              redirects.push(...versionRedirects);
+
               for (const redirect of redirects) {
                 if (existingPath.includes(redirect.to)) {
                   return existingPath.replace(redirect.to, redirect.from);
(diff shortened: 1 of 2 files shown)


Dr-Electron:
So now you could either use:
https://iota-wiki-98xis31i6-iota1.vercel.app/identity.rs/introduction/
or
https://iota-wiki-98xis31i6-iota1.vercel.app/identity.rs/0.6/introduction/

which doesn't work in the current wiki:
http://wiki.iota.org/identity.rs/0.6/introduction/

jlvandenhout:
We can do this later but: have you considered doing it the other way around? So just give every version a `/<project>/<version>` path and then point `/<project>` to a configured default version?

Dr-Electron:
> We can do this later but: have you considered doing it the other way around? So just give every version a `/<project>/<version>` path and then point `/<project>` to a configured default version?

I thought about it. But never went through the idea properly 🤔 
Could make more sense. Let's revisit after we have versioning where we want it to be, where we have redirects for Shimmer and the first verion in the config for every network is the main version

jlvandenhout (approved):

eike-hass:
> Member

The result, or more accurately the deep-linking capabilities would be the same then, right?

Dr-Electron:
> > I thought about it. But never went through the idea properly 🤔
> > Could make more sense. Let's revisit after we have versioning where we want it to be, where we have redirects for Shimmer and the first verion in the config for every network is the main version
> 
> The result, or more accurately the deep-linking capabilities would be the same then, right?

Yes

PR #1216, opened 2023-10-05
Remove goshimmer from the docs

muXxer:
Let's get rid of it in the docs.


Dr-Electron (dismissed):
LGTM, but we should check if we want to recycle some of those docs later on. They are actually nicely written

lucas-tortora (approved):

lucas-tortora:
> LGTM, but we should check if we want to recycle some of those docs later on. They are actually nicely written

GIT keeps everything, and the PR name is clear and easy to track. So that shouldn't be a problem :)

Dr-Electron (approved):

PR #1217, opened 2023-10-05
Remove private tangle from the wiki

muXxer:
I think the changed links are wrong, but you get the idea :)

We don't want to officially support private tangles.
Instead the usage of the testnet should be encouraged.


eike-hass:
I might be late to the party here, but I think the removal of private tangle docs and support is a mistake. I assume serious projects want to be able to spin up private versions of current (or future or even past) protocol versions to validate their applications, node plugins etc. Even if applications are developed against test nets, I would argue using a private tangle in CI is faster and more reliable and should be considered a valid use case.

luca-moser:
I agree with Eike here, we must absolutely provide a means for external parties to be able to set up their own network for testing.

Dr-Electron:
I actually discussed this with @jlvandenhout a while ago. We should commit to create a better ganache like setup. Not just for L2. It should have profiles for INX plugin development, explorer development and so on. This tool could be a really nice tool for any IOTA dev. 
But we also wanted to not call it private tangle as that always comes with those people that actually want to deploy a private tangle in production with sometimes really weird use cases. So any name suggestions?
(Take the second paragraph with some humour, but the first is something I strongly believe in. It could foster contribution)

luca-moser:
Adding a new dropdown item under L1 called "Local Development Network" with a description of `Setup a local network for development/testing.` and a network icon should be sufficient.

PR #1218, opened 2023-10-05
Remove hornet chrysalis docs

muXxer:
I guess that will break other links, but you get the idea :)

IMO we should not provide confusing deprecated docs in the wiki, especially when they are broken, but @luca-moser  told me to rather add a banner that it is deprecated. I guess that is something we should discuss in general.


lucas-tortora:
> I guess that will break other links, but you get the idea :)
> 
> IMO we should not provide confusing deprecated docs in the wiki, especially when they are broken, but Luca told me to rather add a banner that it is deprecated. I guess that is something we should discuss in general.

For the time being, I think we need to still have these docs somewhere with a clear `Deprecated` sign. However, that place doesn't necessarily have to be the main wiki. I was thinking we could create a new domain `legacy.wiki.iota.org` or `deprecated.wiki.iota.org`, where docs go to die basically. WDYT?

muXxer:
For me that would be a super nice solution. Is that possible? What do you say @luca-moser ?

luca-moser:
> > I guess that will break other links, but you get the idea :)
> > IMO we should not provide confusing deprecated docs in the wiki, especially when they are broken, but Luca told me to rather add a banner that it is deprecated. I guess that is something we should discuss in general.
> 
> For the time being, I think we need to still have these docs somewhere with a clear `Deprecated` sign. However, that place doesn't necessarily have to be the main wiki. I was thinking we could create a new domain `legacy.wiki.iota.org` or `deprecated.wiki.iota.org`, where docs go to die basically. WDYT?

Sounds ok to me

Dr-Electron:
That means I can close this for now?

lucas-tortora:
closed in favour of https://github.com/iotaledger/devx/issues/248

PR #1226, opened 2023-10-07
Fix redirects

Dr-Electron:
# Description of change

Fixed multiple bugs in redirects:
- Hornet redirects were not correctly created, because for `to: /hornet/2.0` no docs exist so no redirects were created. So I changed it to `to: /hornet`.
- Only the first redirect was returned so now we return an array
- Instead of `includes` I now used `startsWith` to not replace weird occurrences in the middle.

## Type of change

## Change checklist

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -323,7 +323,7 @@ module.exports = async () => {
                 },
                 {
                   from: '/shimmer/hornet',
-                  to: '/hornet/2.0',
+                  to: '/hornet',
                 },
                 {
                   from: '/shimmer/identity.rs',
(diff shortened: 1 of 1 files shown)


lucas-tortora:
Looks good in theory, though we can't really test this locally. Btw, if it doesn't work I think we should take [Docusaurus's advice](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects) and ask dev ops for server-side redirects.

Dr-Electron:
> Looks good in theory, though we can't really test this locally. Btw, if it doesn't work I think we should take [Docusaurus's advice](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects) and ask dev ops for server-side redirects.

You can test locally. 🤔 
You just need to build and serve 🤔 


We should definitely try to use server side redirects, but not for automated redirects. That's a nice thing for client redirects imo

Dr-Electron:
https://iota-wiki-g13rkwnfn-iota1.vercel.app/

lucas-tortora (approved):
Redirects for dirs don't seem to be working properly, but doc to doc have improved :)

Dr-Electron:
> Redirects for dirs don't seem to be working properly, but doc to doc have improved :)

Can you open an issue describing what doesn't work? 😓

PR #1327, opened 2023-11-07
Use BPS instead of TPS

Dr-Electron:
# Description of change

Replaced TPS with BPS

## Type of change

## Change checklist


marc2332 (changes requested):
Also change the `EpochReward` type as it currently uses `totalTps`

Dr-Electron:
> Also change the EpochReward type as it currently uses totalTps

Not sure what you mean with that 😅 Where does it use totalTps?

marc2332:
> > Also change the EpochReward type as it currently uses totalTps
> 
> Not sure what you mean with that 😅 Where does it use totalTps?

Ah yes, I just remembered that for some reason it no longer has the `totalTps` field, but it should. Let's add it as `totalBps`.

image
image

src/components/HomeLayout/styles.css
```diff
@@ -682,3 +682,66 @@
...
+
+.mana-calculator__button {
+  background-color: var(--ifm-color-emphasis-100);
+  margin: 0px;
+  margin-top: 1rem;
+}
+
+.mana-calculator__tooltip > p {
+  margin: 0;
+}
```
marc2332, line 747: I think you got these duplicated, just remove them from here as I think they should be in ManaCalculator/styles.css

marc2332 (changes requested):

src/components/HomeLayout/styles.css
Dr-Electron, line 747:
```suggestion
```

marc2332 (approved):

PR #1372, opened 2023-12-02
Update docusaurus to 2.4.3

Dr-Electron:
# Description of change

Updated docusaurus to v2.4.3

## Type of change

- Enhancement (a non-breaking change which adds functionality)

## Change checklist

- [x] I have followed the [contribution guidelines](https://github.com/iota-wiki/iota-wiki/blob/main/.github/CONTRIBUTING.md) for this project
- [x] I have performed a self-review of my own changes
- [x] I have made sure that added/changed links still work
- [ ] I have commented my code, particularly in hard-to-understand areas

PR #1373, opened 2023-12-02
Add visual regression testing

Dr-Electron:
# Description of change

Added visual regression testing as recommended by Docusaurus.
Needs to be merged after #1372 

- [x] Which sides do we want to track?
    - Mana Calculator
    - Glossary
    - Network Endpoints (Metamask Button)
    - Create Mnemonic (Code import)
    - Apis (Well, to check the APIs 😆 ) 
- [x] When do we want to trigger it? Because I think it doesn't have to be run for every PR 🤔 Maybe only for PRs with a specific tag?
    - Currently we trigger with every commit on main. And on PRs that have the `Argos` label to compare with main. Currently we take 33 screenshots every time

## Type of change

## Change checklist


playwright.config.ts
```diff
@@ -0,0 +1,19 @@
...
+    {
+      name: 'chromium',
+      use: {
+        ...devices['Desktop Chrome'],
+      },
+    },
+  ],
+};
+
+export default config;
```
Dr-Electron, line 19:
```suggestion
export default config;

```

screenshot.css
```diff
@@ -0,0 +1,19 @@
...
+[class*='playgroundPreview'] {
+  visibility: hidden;
+}
+
+/* Different docs last-update dates can alter layout */
+.theme-last-updated,
+/* Mermaid diagrams are rendered client-side and produce layout shifts */
+.docusaurus-mermaid-container {
+  display: none;
+}
```
Dr-Electron, line 19:
```suggestion
}

```

screenshot.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      'mana-calculator|' +
+      'learn/glossary|' +
+      'build/networks-endpoints|' +
+      'create-mnemonic|' +
+      '(api|apis)/core'
+    ),
+  );
+  console.log('Pathnames to screenshot:', pathnames);
+  pathnames.forEach(screenshotPathname);
+});
```
Dr-Electron, line 41:
```suggestion
});

```

utils.ts
```diff
@@ -0,0 +1,19 @@
...
+  $('loc').each(function handleLoc() {
+    urls.push($(this).text());
+  });
+  return urls.map((url) => new URL(url).pathname);
+}
+
+// Converts a pathname to a decent screenshot name
+export function pathnameToArgosName(pathname: string): string {
+  return pathname.replace(/^\/|\/$/g, '') || 'index';
+}
```
Dr-Electron, line 19:
```suggestion
}

```

screenshot.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+test.describe('Docusaurus site screenshots', () => {
+  const pathnames = extractSitemapPathnames(sitemapPath).filter((pathname) =>
+    pathname.match(
+      ' ^/$|' +
+        'mana-calculator|' +
+        'learn/glossary|' +
+        'build/networks-endpoints|' +
+        'create-mnemonic|' +
+        '(api|apis)/core',
```
jlvandenhout, line 37: This results in quite a few redundant `api` matches. I guess we can just limit this to one page? Like just `/apis/core/v2/returns-the-health-of-the-node` which in principle will always be there, regardless of spec changes?

jlvandenhout: I'm fine either way, it's just that if one thing changes in the API plugin, we need to approve multiple pages for the same change. But I also understand your point, so I'm fine merging it like it is so we can update to V3, and then I'm happy to create a dog feeding page or section in the near future so we don't have this problem. It's only a temporary inconvenience that way.

jlvandenhout (approved):

PR #1539, opened 2024-04-29
Reindex on Algolia crawl

### LINKED ISSUE #1349: Crawl Step in Deploy Action Does Not Trigger a Crawl
**Describe the bug**

The crawl step in the deploy action doesn't trigger a Crawl for the IOTA index in Algolia.

**To Reproduce**

Run the deploy action,  and look at the latest crawl time in https://crawler.algolia.com/admin/crawlers/bd7ad9be-f639-4213-8e62-ccdf4d422136/overview. 

**Expected behavior**
The Algolia crawler should be triggered every single deploy so the search indexes are updated.

**Additional context**
It probably has something to do with the Index ID or API Keys.

Dr-Electron:
# Description of change

This PR should force a reindex after the crawler run. I was too lazy to check docs so I just stole from the official [GitHub action ](https://github.com/algolia/algoliasearch-crawler-github-actions/blob/bdf5357c4a4cc49ea728d7d7207acb454113d10a/src/crawler-api-client.ts#L302-L305) :trollface: 
We could also use the GitHub action. But I think in our case it isn't really needed.

Fixes #1349.

## Type of change

## Change checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,7 +42,5 @@ jobs:
         run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
       # Trigger Algolia crawler
       - name: Crawl
-        uses: algolia/algoliasearch-crawler-github-actions@v1
         run: |
-          curl -H "Content-Type: application/json" -X POST --user ${{ secrets.ALGOLIA_USER_ID }}:${{ secrets.ALGOLIA_API_KEY }} "https://crawler.algolia.com/api/1/crawlers/${{ secrets.ALGOLIA_CRAWLER_ID }}/run"
           curl -H "Content-Type: application/json" -X POST --user ${{ secrets.ALGOLIA_USER_ID }}:${{ secrets.ALGOLIA_API_KEY }} "https://crawler.algolia.com/api/1/crawlers/${{ secrets.ALGOLIA_CRAWLER_ID }}/reindex"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,5 +42,7 @@ jobs:
         run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
       # Trigger Algolia crawler
       - name: Crawl
+        uses: algolia/algoliasearch-crawler-github-actions@v1
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -42,5 +42,7 @@ jobs:
         run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
       # Trigger Algolia crawler
       - name: Crawl
+        uses: algolia/algoliasearch-crawler-github-actions@v1
```
jlvandenhout, line 45: Is this really needed when we use the REST API?

.github/workflows/deploy.yml
```diff
@@ -42,5 +42,7 @@ jobs:
...
         run: |
           curl -H "Content-Type: application/json" -X POST --user ${{ secrets.ALGOLIA_USER_ID }}:${{ secrets.ALGOLIA_API_KEY }} "https://crawler.algolia.com/api/1/crawlers/${{ secrets.ALGOLIA_CRAWLER_ID }}/run"
```
jlvandenhout, line 47: We can remove this. From the docs I understand it only activates the crawler, it doesn't actually trigger a crawl till the next schedule. The below command should actually trigger a crawl.

.github/workflows/deploy.yml
Dr-Electron: haha, no. Seems like I just forgot it after trying it out 😅 
```suggestion
```

.github/workflows/deploy.yml
```diff
@@ -42,5 +42,7 @@ jobs:
...
+          curl -H "Content-Type: application/json" -X POST --user ${{ secrets.ALGOLIA_USER_ID }}:${{ secrets.ALGOLIA_API_KEY }} "https://crawler.algolia.com/api/1/crawlers/${{ secrets.ALGOLIA_CRAWLER_ID }}/reindex"
```
Dr-Electron, line 48:
```suggestion
          curl -H "Content-Type: application/json" -X POST --user ${{ secrets.ALGOLIA_USER_ID }}:${{ secrets.ALGOLIA_API_KEY }} "https://crawler.algolia.com/api/1/crawlers/${{ secrets.ALGOLIA_CRAWLER_ID }}/reindex"

```

.github/workflows/deploy.yml
Dr-Electron: Yeah, seems correct 🤔 
```suggestion
```

jlvandenhout (approved):

PR #1560, opened 2024-05-15
Implementation of filters in custom search modal.

vivekjain23:
# Description of change

* Added facet filter on search bar.

fixes https://github.com/iotaledger/iota/issues/1438

## Type of change

## Change checklist


src/theme/DocItem/Layout/index.tsx
Dr-Electron, line 1: Why did this get removed?

src/theme/Footer/index.tsx
Dr-Electron, line 1: Do we need to touch this file?
And please don't remove the commend at the top. It helps us keep track of [swizzled](https://docusaurus.io/docs/swizzling) versions of Docusaurus components

src/utils/searchConstant.ts
Dr-Electron, line 1: This seems horrible to maintain. Can't we import this from configs or something similar?

Dr-Electron, line 1: This seems horrible to maintain. Can't we import this from configs or something similar?

src/theme/Footer/index.tsx
Dr-Electron, line 1: Do we need to touch this file?
And please don't remove the commend at the top. It helps us keep track of [swizzled](https://docusaurus.io/docs/swizzling) versions of Docusaurus components

src/theme/DocItem/Layout/index.tsx
Dr-Electron, line 1: Why did this get removed?

Dr-Electron (changes requested):

src/utils/searchConstant.ts
vivekjain23: I think while updating the filter with instantSearch we can get rid of this.

src/theme/Footer/index.tsx
vivekjain23: It was needed to add few styles in the footer since while styling and alinging the filter, footer's position was getting disturbed.

src/theme/DocItem/Layout/index.tsx
vivekjain23: Was deleted by mistake, adding again. 
Thanks a lot for reminding!

lucas-tortora:
It seems we're having some css alignment issues. If you need a hand with this please let me know :) 

image

image

lucas-tortora:
@Ginowine  and @salaheldinsoliman , could you please review this?

lucas-tortora:
It's looking better, but we still have some issues:

Please fix the following:

1. When you search within an open scrolled-down page we still have alignment issues:

image

2. The searched term is deleted when I enable or disable a filter using the navbar search. This bug doesn't seem to happen when using the search bar on the `More results` page. (see attached video)

[Screencast from 2024-06-12 12-38-45.webm](https://github.com/iotaledger/iota-wiki/assets/85233773/59d333bc-19c3-4d29-9cb5-024a8c9cf31c)

salaheldinsoliman:
The bar on the bottom side looks a bit off when zooming in/out: 
image

I don't know if that's only in my browser though, but it looks normal in the current Wiki

lucas-tortora:
@vivekjain23 , this popped up on the build test https://github.com/iotaledger/iota-wiki/actions/runs/9519155072/job/26241722402?pr=1560#step:7:159

lucas-tortora:
@jlvandenhout, @Dr-Electron, could you take a look at the build error we're getting?  I haven't been able to point out it's trigger this time

Dr-Electron:
So I'm not going to pretend I understand any js code. But I can give you the actually error output without the bold error:

```
'[ERROR]' Docusaurus server-side rendering could not render static page with path /search/.
'[INFO]' It looks like you are using code that should run on the client-side only.
To get around it, try using `<BrowserOnly>` (https://docusaurus.io/docs/docusaurus-core/#browseronly) or `ExecutionEnvironment` (https://docusaurus.io/docs/docusaurus-core/#executionenvironment).
It might also require to wrap your client code in `useEffect` hook and/or import a third-party library dynamically (if any).


ReferenceError: window is not defined
```

src/theme/Footer/index.tsx
```diff
@@ -1,22 +1,20 @@
-/**
- * SWIZZLED VERSION: 2.0.0-rc.1
- * REASONS:
- *  - The socials bar needed full width, so needed to be placed outside the default container.
- */
-
```
jlvandenhout, line 6: Can you please re-add this, amended with appropriate reasons?

src/theme/Footer/index.tsx
```diff
@@ -1,22 +1,20 @@
...
-    <>
-      <Footer {...props} />
-      <Social />
-    </>
+    <div>
+      <div style={footerStyleProps}>
+        <Footer />
+        <Social />
+      </div>
+    </div>
```
jlvandenhout, line 16: Why the nested divs?

src/theme/Layout/Provider/index.js
jlvandenhout, line 1: Please add a header with the version you swizzled and the reasons why.

src/theme/Layout/index.js
jlvandenhout, line 1: Please add a header with the version you swizzled and the reasons why.

src/theme/Root.js
jlvandenhout, line 1: Please add a header with the version you swizzled and the reasons why.

src/theme/SearchBar/index.js
jlvandenhout, line 1: Please add a header with the version you swizzled and the reasons why.

src/theme/SearchPage/index.js
jlvandenhout, line 1: Please add a header with the version you swizzled and the reasons why.

src/theme/SearchBar/FilterDropdown.js
```diff
@@ -0,0 +1,168 @@
...
+    }, [selectedFacets, setSelectedFacets]);
+
+    useEffect(() => {
+      const handleScroll = () => {
+        if (window.scrollY > 28) {
+          setIsScrolled(true);
+        }
+      };
+      handleScroll();
+    }, [window.scrollY]);
```
jlvandenhout, line 107: I think this is the source of the build failure. This dropdown is used on the search page, which means it will be rendered during static site generation. `window.scrollY` is used here as a dependency, but `window` is only available client-side, not server-side. See here how to resolve this: https://docusaurus.io/docs/advanced/ssg#escape-hatches

vivekjain23: Thank you for the help above link, this issue is now resolved.

lucas-tortora:
> The search query is cleared when the filter is changed in the search modal. This differs from the behavior on the search page, where it just updates the search results. I'd prefer if we can get the search modal to behave the same as the search page.

@vivekjain23 , could you look into this, please? It's been 2 weeks since this comment

jlvandenhout:
I have tried to solve it, but my React knowledge is too limited, unfortunately.

jlvandenhout:
After using the filter for a while, I also think we should make the selection behavior a bit different. Have the "all" filter still select everything, but as soon as you select individual items, I think it is more convenient to clear everything except the one selected item on the first individual selection.

Something like:

All is selected:
```
x All
x A
x B
x C
```

User selects B:
```
  All
  A
x B
  C
```

User selects A too:
```
  All
x A
x B
  C
```

vivekjain23:
> After using the filter for a while, I also think we should make the selection behavior a bit different. Have the "all" filter still select everything, but as soon as you select individual items, I think it is more convenient to clear everything except the one selected item on the first individual selection.
> 
> Something like:
> 
> All is selected:
> 
> ```
> x All
> x A
> x B
> x C
> ```
> 
> User selects B:
> 
> ```
>   All
>   A
> x B
>   C
> ```
> 
> User selects A too:
> 
> ```
>   All
> x A
> x B
>   C
> ```

Yupp, thanks for the suggestion. I have implemented it

lucas-tortora (approved):

lucas-tortora (approved):

src/theme/Footer/index.tsx
```diff
@@ -19,4 +19,4 @@ export default function FooterWrapper(props: Props): JSX.Element {
       <Social />
     </>
   );
-}
+}
```
Dr-Electron, line 22:
```suggestion
}

```

src/theme/Root.js
Dr-Electron, line 1: Do we want this? I think in most cases the search filter state should revert again 🤔

src/theme/SearchPage/FilterDropdown.js
Dr-Electron, line 1: Can you make this tsx? Will make it easier for us to migrate to the new docs

src/theme/SearchPage/index.js
Dr-Electron, line 1: Can you swizzle this in ts please?

src/utils/searchConstant.ts
Dr-Electron, line 1: Can't we automate this somehow? We will never maintain it

Dr-Electron (changes requested):

Dr-Electron:
Btw there is also still a bug, if there is only 1 search result for example, the footer is not at the bottom of the page

Dr-Electron:
Let's close this and potentially revisit on the new docs

